PR #53, opened 2019-02-25
Feat/multiple storyboards

havard024:


ostelco-ios-client/src/MainController.swift
```diff
@@ -11,56 +11,79 @@ import UIKit
 class MainController: UIViewController {
     
     @IBAction func unwindFromLoginViewController(sender: UIStoryboardSegue) {
-
+        perform(#selector(showSignUp), with: nil, afterDelay: 0)
```
havard024: Without a delay, you end up with this error message in the unwind function:

2019-02-25 18:48:24.060175+0100 dev-ostelco-ios-client[43694:475464] Warning: Attempt to present <dev_ostelco_ios_client_app.SignUpViewController: 0x7f8283514980> on <dev_ostelco_ios_client_app.MainController: 0x7f828050d060> whose view is not in the window hierarchy!

PR #104, opened 2019-04-05
Feature/cleanup code

prasanthu:


ostelco-ios-client/src/extensions/UIViewController+UIActivityIndicator.swift
```diff
@@ -8,32 +8,27 @@
...
-        
-        vSpinner = spinnerView
-        
-        return vSpinner
+        return spinnerView
     }
-    
-    func removeSpinner() {
+
+    func removeSpinner(_ spinnerView: UIView?) {
```
havard024: I think it would make sense to make the spinnerView parameter mandatory. Since the global spinner variable is gone, you would only call the removeSpinner function if you actually want to remove a spinner.

PR #106, opened 2019-04-07
Feature/use promise

prasanthu:


ostelco-ios-client/src/features/splash/SplashViewController.swift
```diff
@@ -26,208 +26,155 @@ class SplashViewController: UIViewController {
...
+
+    func getFullfilledPromise<T>() -> Promise<T?> {
+        let empty:T? = nil
+        return Promise<T?> { seal in
+            seal.resolve(empty, nil)
         }
-        OnBoardingManager.sharedInstance.region = region
+    }
+
+    func getRejectedContext() -> Promise<Context?> {
```
havard024: This function seems to be the same as as the `getContext` on line 55 and doesn't seem to be used in this file

ostelco-ios-client/src/features/splash/SplashViewController.swift
```diff
@@ -26,208 +26,155 @@ class SplashViewController: UIViewController {
...
+                    } else {
+                        rejectPromise(error)
+                    }
+                }.onCompletion { _ in
+                    self.removeSpinner(self.spinnerView)
+            }
+        }
+    }
+
+    func getSegueFromRegionResponse(region: RegionResponse) -> String {
```
havard024: [LoginViewController2](https://github.com/ostelco/ostelco-ios-client/blob/feature/use-promise/ostelco-ios-client/src/features/login/LoginViewController2.swift ) could also benefit from this function, it has similar logic to the splash screen, and uses the exact same logic for routing and setting credentials after login is complete

havard024: Except for how it performs the segue: 
```
self.perform(#selector(self.showEKYCLastScreen), with: nil, afterDelay: 0.5)
```
For some reason in needs the delay, else the segue won't show, I assume this is an logic error from my part

designatednerd:
@prasanthu You wanna merge and/or rebase off master and I'll take a look?

prasanthu:
> @prasanthu You wanna merge and/or rebase off master and I'll take a look?

@designatednerd This was my 1st attempt using PromiseKit, not satisfied with this. Will create a new one later

designatednerd:
OK - sounds like a plan. I can try to do an example API if that'd help!

PR #107, opened 2019-04-08
Update Pods structure + start warning cleanup

designatednerd:
In this PR: 

- Updated the structure of the `Podfile` so we're only declaring dependencies in one place (and where possible, specifying version)
- Updated base `xcconfig` for all targets to use podfile ones - Note that this will break local builds for a bit until we get the secrets stuff squared away
- Removed pods we're not actively using
- Added PromiseKit current version. 
- Deleted unused code in `old-view-controllers`
- Deleted unused VC's in a couple storyboards to start dealing with warnings (most of the rest of this will come in a separate PR)


designatednerd:
[Use abstract target in Podfile](https://trello.com/c/x9o2G66i/17-use-abstract-target-in-podfile)

PR #108, opened 2019-04-08
Update Folder Structure

designatednerd:
In this PR: 

- Update folder structure to be more like other iOS projects. Please let me know if you have questions about the updated structure, but at the top level, it's a bit more divided by "What is this" rather than "What specific bit of the app is this for?". `ViewControllers` subfolders are set up to align more with "What specific bit of the app this for?". The idea is to try and encourage *some* reuse where possible.
- Remove a bunch of stuff that's not being used anymore
- Break up gargantuan `UIHelpers` into files per class/extension
- Clean up existing storyboards and break them apart per tab
- Add Test App Delegate to facilitate running tests without UI, and one actual test
- Added a couple sample classes for new folders in the updated structure
- Killed off all warnings which were not intentionally added.


ostelco-ios-client/TestAppDelegate.swift
```diff
@@ -0,0 +1,34 @@
...
+//  TestAppDelegate.swift
+//  ostelco-ios-client
+//
+//  Created by Ellen Shapiro on 4/8/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import UIKit
+
+class TestAppDelegate: UIResponder, UIApplicationDelegate {
```
designatednerd: All this goop is based on this technique: https://qualitycoding.org/ios-app-delegate-testing/

ostelco-ios-client/Models/Context.swift
```diff
@@ -19,7 +19,9 @@ struct Context: Codable {
     func getRegion() -> RegionResponse? {
 
         var ret: RegionResponse? = nil
-        var hasPendingStatus = false
+        
+        #warning("Figure out what this is for before deleting it entirely")
```
designatednerd: @prasanthu @havard024 Any idea what this `hasPendingStatus` boolean is for?

ostelco-ios-client/ModelControllers/UserDefaultsWrapper.swift
```diff
@@ -0,0 +1,60 @@
...
+
+struct UserDefaultsWrapper {
+    
+    // Underlying keys for user defaults
+    private enum Key: String, CaseIterable {
+        case selectedTheme = "SelectedTheme"
+    }
+    
+    private static var defaults: UserDefaults {
+        return UserDefaults.standard
```
designatednerd: This is set up as a wrapper because it's possible that to support some stuff in an extension, we may need to move this to a framework and use a shared user defaults suite.

Also it makes things cleaner at the call site.

ostelco-ios-client/ThemeManager.swift
```diff
@@ -1,73 +0,0 @@
...
-            return Theme(rawValue: storedTheme)!
-        } else {
-            return .TurquoiseTheme
-        }
-    }
-    
-    static func applyTheme(theme: Theme) {
-        // First persist the selected theme using NSUserDefaults.
-        UserDefaults.standard.setValue(theme.rawValue, forKey: SelectedThemeKey)
-        UserDefaults.standard.synchronize()
```
designatednerd: FYI if you're ever working on another iOS project: The [docs for `NSUserDefaults.synchronize`](https://developer.apple.com/documentation/foundation/nsuserdefaults/1414005-synchronize?language=objc) say that "this method is unnecessary and shouldn't be used."

Code level comments offer suggestions for migrating to other methods and/or just removing the call altogether. Can't link to the source but someone helpfully copied them [here](https://stackoverflow.com/a/47648278/681493).

ostelco-ios-client/ViewControllers/Country/AllowLocationAccessViewController.swift
```diff
@@ -91,7 +91,8 @@ class AllowLocationAccessViewController: UIViewController {
                 showLocationAccessRestricted()
             case .denied:
                 showLocationAccessDenied()
-            case .authorizedAlways, .authorizedWhenInUse:
+            case .authorizedAlways,
+                 .authorizedWhenInUse:
```
designatednerd: for what it's worth, I find it clearer to put these on separate lines for a couple reasons: 

a) Makes it way more obvious that there is more than one case being handled
b) Makes diffs clearer when additional cases are added or removed to this same handling block.

ostelco-ios-client/ViewControllers/Country/LocationAccessDeniedViewController.swift
```diff
@@ -29,10 +29,16 @@ class LocationAccessDeniedViewController: UIViewController {
...
-                case .restricted, .denied:
+                case .restricted,
+                     .denied:
                     return false
-                case .authorizedAlways, .authorizedWhenInUse, .notDetermined:
+                case .authorizedAlways,
+                     .authorizedWhenInUse,
+                     .notDetermined:
                     return true
+                @unknown default:
```
designatednerd: This is a new thing in Swift 5 that essentially allows you to add a default case for enums you don't control and which might in the future have other values. 

If you're extremely bored [please enjoy the Swift Evolution proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md) which led to this change.

ostelco-ios-client/Models/Context.swift
prasanthu: Lets remove it.

havard024: Should be fine to remove it.
Some back story of this function, our app assumes that you only belong to one region while the APIs support multiple. Thus the APIs return a list and the app picks out the correct region, this function could easily just picked out the first region in the list

designatednerd:
@prasanthu Just using `feature/blah` for everything rather than `update/blah` etc? S'ok!

designatednerd:
[Update folder structure to be more aligned with iOS style](https://trello.com/c/glvmZzG5/18-update-folder-structure-to-be-more-aligned-with-ios-style)

designatednerd:
[Kill warnings with fire](https://trello.com/c/55Ojn6GK/19-kill-warnings-with-fire)

PR #109, opened 2019-04-08
Fix the build settings

prasanthu:
This will make sure we use  Development.xcconfig & Production.xcconfig files

### DIFF
--- a/Pods/Pods.xcodeproj/project.pbxproj
+++ b/Pods/Pods.xcodeproj/project.pbxproj
@@ -5476,13 +5476,14 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0930;
-				LastUpgradeCheck = 1020;
+				LastUpgradeCheck = 0930;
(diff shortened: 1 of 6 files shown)


Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -7873,7 +7872,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
-				CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED = YES;
```
designatednerd: Why did all the stuff in this file get reverted?

prasanthu: My Xcode is doing this : Version 10.2 (10E125)

prasanthu:
> I've got something else in the works that should remove the need for these - do you want to do this as a temporary fix?

No need for temporary fix.

PR #110, opened 2019-04-10
Script secret hiding with Swift Package Manager

designatednerd:
Apologies for the length on this one. Though a good chunk of it is Xcodeproj garbage, I promise I'll get these down to WAY more reasonable sizes going forward. 

In this PR: 

- Added a SPM-based project which can be run from the command line that will use a hidden JSON file to inject either prod or dev values during the build process, then reset those values back to garbage after the build completes. 
- Added pre- and post- build phase for this on both prod and dev targets
- Moved several keys from the `Info.plist` file into their own `Environment.plist` file to make them easier to adjust per target
- Added tests on SPM scripts to make sure everything is being updated properly and/or is failing properly when insufficient information is provided
- Removed existing methods of hiding secrets
- Deleted var whose deletion was approved in #108.

TODO in next PR: 
- Update README to explain all this
- Update Bootstrap scripts
- Update tests on CI to run 
- Add buckets more documentation to classes


.circleci/config.yml
```diff
@@ -12,14 +12,23 @@ jobs:
     shell: /bin/bash --login -o pipefail
     steps:
       - checkout
+      - restore_cache:
+          key: "v2-gems"
       - run:
-          name: Fetch CocoaPods Specs
-          command: |
-            curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
+          name: Install with bundler
```
prasanthu: I got this sample from https://circleci.com/docs/2.0/ios-migrating-from-1-2/#example-20-ios-project-configuration

scripts/SPM/Sources/Core/File+Reset.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+
+import Foundation
+import Files
+import ShellOut
+
+extension File {
+    
+    func resetToGitHEAD() throws {
+        let cmd = ShellOutCommand(string: "git checkout HEAD -- \"\(self.path)\"")
```
prasanthu: There might be a problem when you are in a detached HEAD state in git (e.g checking out a tag).  See the circle ci workflows

scripts/SPM/Sources/Core/Secrets/EnvironmentUpdater.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Basic
+//
+//  Created by Ellen Shapiro on 4/9/19.
+//
+
+import Foundation
+import Files
+
+struct EnvironmentUpdater {
+    enum EnvironmentKey: String, CaseIterable, KeyToUpdate {
```
prasanthu: This needs proper documentation. Adding a new entry is not obvious.

scripts/SPM/Sources/Core/File+Reset.swift
prasanthu: Also will this revert any changes made to the project file when building using fastlane? Our `tfbeta` lane updates the build number.

.circleci/config.yml
designatednerd: Yeah, I definitely am seeing the downsides of using the `pod outdated` method, I'll probably put this back if I can't find a better way to cache the checkout

scripts/SPM/Sources/Core/File+Reset.swift
designatednerd: It will only check out `HEAD` for the single file represented by the `file` object. It will not affect other files.

scripts/SPM/Sources/Core/Secrets/EnvironmentUpdater.swift
designatednerd: Wholeheartedly agreed, which is why this is in the list of TODOs in the PR for the *next* PR.  

I wanted to get this shipped first so building properly could happen while I wrote up the docs.

PR #114, opened 2019-04-11
Feature/minor fixes to make testing easier

havard024:
Added feature to open need help action sheet by shaking phone for cases where we have forgotten or don't have an option to open that menu. This makes in possible for us to open the in app chat or to start again completely in case you are stuck somewhere.

need help in EKYC pending verification screen was a label, replaced it with a button that opens need help action sheet

show spinner when clicking retry on EKYC pending verification screen

The region in OnBoardingManager is only set when we call context from splash screen or after you log in, for new users the RegionResponseArray inside context will be empty thus the region in OnBoardingManager is never set and this caused a crash when trying to use the OnBoardingManager region to fetch latest simProfiles on the esim pending download screen

Added code to always send you to EKYC on boarding even if location + selected country verification fails. So now everyone will be treated as they are in singapore.


ostelco-ios-client/Models/Context.swift
```diff
@@ -8,6 +8,35 @@
 
 // TODO: Use Region struct from Region.swift when feature/ekyc-api-integration branch is merged into develop
 
+func getRegionFromRegionResponseArray(_ regionResponses: [RegionResponse]) -> RegionResponse? {
```
havard024: This has now become a helper function, should it be placed in a different file than the model file itself?

ostelco-ios-client/Network/APIManager.swift
```diff
@@ -71,3 +71,34 @@ class APIManager: Service {
         }
     }
 }
+
+extension APIManager {
+    
+    #warning("NSError should contain useful information inside onSuccess")
+    func getRegionFromRegions(completion: @escaping (RegionResponse?, Error?) -> Void) {
```
havard024: Thought we could refactor how we use the APIManager, because everytime we use any resource we need to duplicate the below code.

ostelco-ios-client/Network/APIManager.swift
```diff
@@ -71,3 +71,34 @@ class APIManager: Service {
         }
     }
 }
+
+extension APIManager {
+    
+    #warning("NSError should contain useful information inside onSuccess")
```
havard024: Not sure how we should handle the errors inside the onSuccess callback when there is no error to return

ostelco-ios-client/ViewControllers/ESim/ESIMPendingDownloadViewController.swift
```diff
@@ -113,4 +111,20 @@ class ESIMPendingDownloadViewController: UIViewController {
             }
         }
     }
+    
+    override func viewDidLoad() {
+        if let region = OnBoardingManager.sharedInstance.region {
+            getSimProfileForRegion(region: region)
+        } else {
+            #warning("RegionResponse contains simProfile so we don't need to fetch the simProfiles for region in this case.")
```
havard024: I'll fix this right now and push and update to this PR

ostelco-ios-client/Models/Context.swift
prasanthu: @designatednerd Was using `class func` methods to do this

ostelco-ios-client/Network/APIManager.swift
prasanthu: Will you be able to use PromiseKit here?.  The default is always the main queue

ostelco-ios-client/Network/APIManager.swift
prasanthu: If we don't have a response object or a region inside that, is that an error?  If it is to be expected, we can return `nil` for error. The user of the method should expect a `nil` region

ostelco-ios-client/ViewControllers/ESim/ESIMPendingDownloadViewController.swift
prasanthu: ok

ostelco-ios-client/Network/APIManager.swift
havard024: That's one possibility, I was thinking we should construct our own error type, with a descriptive message (maby localizedString?), so that if region is nil, there is always an error and the view does not need to check the error type and decide what text to show, since we decide what to show in the error itself

ostelco-ios-client/ViewControllers/ESim/ESIMPendingDownloadViewController.swift
havard024: fixed, this wasn't an issue, the getSimProfileForRegion already checks region for a simProfile and only calls get /simProfiles if region does not contain any

ostelco-ios-client/ViewControllers/ESim/ESIMPendingDownloadViewController.swift
```diff
@@ -113,4 +111,19 @@ class ESIMPendingDownloadViewController: UIViewController {
             }
         }
     }
+    
+    override func viewDidLoad() {
```
designatednerd: a) Please don't forget to call `super` on view lifecycle methods
b) General iOS convention is to put view lifecycle methods towards the top of the file, under variable declarations.

ostelco-ios-client/Models/Context.swift
havard024: @designatednerd where would you put this function?

ostelco-ios-client/Extensions/UIViewController+ActionSheet.swift
```diff
@@ -25,28 +25,39 @@ class NeedHelpAlertController: UIAlertController {
...
+        }
+    }
+    
+    return alertAction
+}
+
 extension UIViewController {
+    
+    @objc func showSplashScreen() {
+        let viewController = UIStoryboard(name: "Splash", bundle: nil).instantiateInitialViewController()!
```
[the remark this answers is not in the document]
havard024: Great, I saw it, Ill change that

ostelco-ios-client/Extensions/UIViewController+Gesture.swift
```diff
@@ -0,0 +1,23 @@
...
+        if motion == .motionShake {
+            showNeedHelpActionSheet()
+        }
+    }
+
+    override open var canBecomeFirstResponder: Bool {
+        get {
+            return true
+        }
+    }
```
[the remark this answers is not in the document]
havard024: Would that imply changing all our extensions to extend a base view controller, then make all our view controllers inherit from the base? If yes, I'll add a task in trello to restructure

ostelco-ios-client/Network/APIManager.swift
havard024: I'll try to add these errors

ostelco-ios-client/Extensions/UIViewController+Gesture.swift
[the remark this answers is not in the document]
designatednerd: No, only ones where you're trying to override something. I'm actually kind of surprised the compiler isn't whining about this, you shouldn't be able to override functions in an extension

ostelco-ios-client/Network/APIManager.swift
havard024: Mind if I add a warning to add PromiseKit for now? I know it's bad to leave unfinished business, but I want a new test flight version out with these fixes. Right now the one guy in our team who's testing almost everyday is completely stuck 😓 
I'll make sure I fix it afterwards

ostelco-ios-client/ViewControllers/ESim/ESIMPendingDownloadViewController.swift
havard024: I'll fix this

ostelco-ios-client/Network/APIManager.swift
designatednerd: Doesn't need to be a warning, can just be a `// todo`. And this is fine by me

havard024: I'll do that, and do the refactoring when your branch is merged with the APIHelper

havard024: I've added todos to refactor into PromiseKit

ostelco-ios-client/Network/APIManager.swift
havard024: I've added internal error enum inside APIManager for now, until we refactor

ostelco-ios-client/Extensions/UIViewController+Gesture.swift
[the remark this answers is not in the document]
havard024: Now that's interesting, I tried to add this in the AppDelegate, but could not get the gesture to trigger. I want the gesture event to be accessible from anywhere in the app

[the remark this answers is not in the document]
designatednerd: Yeah the app delegate is just an object in the responder chain, other than hanging on to/ setting up the window it doesn't really have any effect on the UI. There's some stuff where you sort of *can* use it as a singleton, but probably shouldn't.

ostelco-ios-client/ViewControllers/SplashViewController.swift
```diff
@@ -8,7 +8,10 @@
 
 import UIKit
 
-class SplashViewController: UIViewController {
+class SplashViewController: UIViewController, StoryboardLoadable {
+    static var storyboard: Storyboard = Storyboard(rawValue: "Splash")!
```
designatednerd: you should be able to just make this `.splash` and not have to force unwrap

ostelco-ios-client/ViewControllers/SplashViewController.swift
```diff
@@ -8,7 +8,10 @@
...
+    static var isInitialViewController = true
```
designatednerd: IIRC you can make this a `static let` since the `var` only requires a `get`

PR #128, opened 2019-04-12
remove auth0 logout and persist userLoggedOut flag

havard024:


ostelco-ios-client/Network/Auth.swift
```diff
@@ -18,25 +18,20 @@ class Auth {
...
         os_log("Clear credentials in auth0 credentials manager.")
-        Auth0
-            .webAuth()
-            .clearSession(federated: true) {
-                print("Clear Session: ", $0)
-                _ = self.credentialsManager.clear()
-                if let callback = callback {
-                    callback()
-                }
+        self.credentialsManager.clear()
```
prasanthu: This will not fully logout a user.  Are you sure ?

havard024: Yes, unless we have a good reason requiring us to fully logout a user.
For me, a good reason is that the users JWT is invalidated, but that's not the case, we can still access prime using a logged out users JWT. Also if we have a short expiry date on JWT, like 10 minutes, they will be logged out in auth0 after that amount of time either way.

There are some more discussions in this card https://trello.com/c/ZkmmJuWl

prasanthu: What is the point of having a logout and it won't do that. If user is a given a chance to logout, it should log out fully.

ostelco-ios-client/ModelControllers/UserDefaultsWrapper.swift
```diff
@@ -57,4 +58,17 @@ struct UserDefaultsWrapper {
             }
         }
     }
+    
+    static var userLoggedOut: Bool {
```
designatednerd: I think this is not necessary - if we don't have anything stored for the user in the keychain, we'll have to kick them  to the login flow anyway to get a new JWT, regardless of whether we manage to invalidate their session with Auth0.

ostelco-ios-client/Network/Auth.swift
designatednerd: This at least clear's the user's token from our application, which forces them to re-login through Auth0. Now granted, if they've still got a cookie set by Auth0, @prasanthu  is correct that "forcing them to relogin" basically means "Show a screen for 2 seconds, then move on," but it's at least better than nothing.

Looking at [Auth0's logout docs](https://auth0.com/docs/logout), phase 2 of "Log users out of Auth0 by clearing the single-sign on cookie" is a good place to start. I'm not 100% sure we have access to this cookie with the type of flow we're using, but if there's a way to access it, nuking it at the cookie storage level is probably going to be better than either just deleting the token or having to show them a webview.

PR #132, opened 2019-04-14
Release/20190414094017

havard024:


havard024:
There's an issue where the app sends the user to the "Oh no something went wrong screen" after they click continue on the eSIM on boarding screen. From what I've experienced, this happens to new users if they finish the on boarding in one sitting. 
Earlier, we got an exception that region in the on boarding manager was not set. The region in on boarding manager is only set when user opens the app or logs in with an existing account, else it is set to nil. This I tried to fix by calling `/regions` inside `ESIMPendingDownloadViewController` if region was nil.
We stopped getting the region force unwrap nil value exception, but an error still occured.
I tried again to fix it here 
d1c88a7 also tried to wrap the fix to run inside the main thread, but still the app sends the user to the error screen.
I'm not getting anything useful out of our logs either.
Next I'll try to use breakpoints to understand what happens inside `ESIMPendingDownloadViewController`

PR #136, opened 2019-04-15
Change Traget name

prasanthu:
This will make sure that the sign-in dialog shows the proper application name. It should show "Oya development app" instead of ~ "dev-ostelco-ios-app"


prasanthu:
We did this another way:  https://github.com/ostelco/ostelco-ios-client/pull/137

PR #137, opened 2019-04-15
Update Product Name to be user-friendlier

designatednerd:
A stab at doing #136 a bit simpler.

Before: 
IMG_1767

After: 
Simulator Screen Shot - iPhone Xʀ - 2019-04-15 at 16 16 54

### DIFF
--- a/dev-ostelco-ios-clientTests/EnumHelperTests.swift
+++ b/dev-ostelco-ios-clientTests/EnumHelperTests.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2019 mac. All rights reserved.
 //
 
-@testable import dev_ostelco_ios_client_app
+@testable import Oya_Development_app
 import XCTest
 
 class EnumHelperTests: XCTestCase {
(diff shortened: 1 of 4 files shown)


designatednerd:
Note: The `-` is there instead of a space because Xcode freaks the hell out if you have a space in your `Product Name`. Probably worth eventually investigating what key Auth0 is using to pull this (they should *probably* be using the bundle display name and not the Product name), but might be overkill since the prod version is just "Oya".

prasanthu:
@oisinzimmermann do you think "Oya-development" is enough ?

designatednerd:
Note that prod version will just by "Oya"

prasanthu:
@designatednerd The way I did it allows space ;-)

designatednerd:
I mean, fair, but I feel like if we don't absolutely need that it's not worth the additional hassle

prasanthu:
@designatednerd I think you have to change Pod names now. If you can't avoid that, it wil most like be very similar to https://github.com/ostelco/ostelco-ios-client/pull/136.

prasanthu:
> @designatednerd I think you have to change Pod names now. If you can't avoid that, it will most like be very similar to #136.

I thought the build failed. May be I clicked the wrong link for circle CI

designatednerd:
@prasanthu With this method you should not have to change the pod names because the names of the targets themselves have not changed. That's a huge part of why this is WAY fewer changes.

PR #138, opened 2019-04-16
Remove Swifty JSON in favor of Codable parsing

designatednerd:
We were using `SwiftyJSON` essentially for one piece of parsing - I've updated this to use `Codable` parsing instead, and added a test against some test data we were already using.

Note: Actually removing the framework will come in another PR, wanted to separate Pod changes from code changes here.

### DIFF
--- a/ostelco-ios-client/Models/MyInfoDetails.swift
+++ b/ostelco-ios-client/Models/MyInfoDetails.swift
@@ -8,7 +8,7 @@
 
 import SwiftyJSON
 
-struct MyInfoAddress{
+struct MyInfoAddress: Codable {
     let country: String?
     let unit: String?
     let street: String?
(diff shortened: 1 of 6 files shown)


ostelco-ios-client/Models/MyInfoDetails.swift
```diff
@@ -17,78 +17,101 @@ struct MyInfoAddress{
...
         return addressLine2
     }
 }
 
-struct MyInfoDetails {
-    let name: String
-    let dob: String
-    let email: String
+struct MyInfoDetails: Codable {
+    private let _name: MyInfoRequiredValue
```
designatednerd: This is using an old Obj-C convention for private variables. The idea here is that we want to mask the underlying details of this to the caller, but still allow the `Codable` parsing system to work with structures we understand.

ostelco-ios-client/Models/MyInfoDetails.swift
```diff
@@ -17,78 +17,101 @@ struct MyInfoAddress{
...
+    var nationality: String? {
+        return _nationality?.value
+    }
+    
     let address: MyInfoAddress
-    let sex: String?
-    let residentialStatus: String?
-    let nationality: String?
-    let mobileNumber: String?
+    let mobileNumber: MyInfoMobileNumber?
```
designatednerd: `address` and `mobileNumber` are directly exposed since there's some underlying details we may need to preserve there.

PR #141, opened 2019-04-16
Add missing framework to dev target

prasanthu:


ostelco-ios-client.xcodeproj/project.pbxproj
```diff
@@ -1802,7 +1816,7 @@
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = "$(inherited)";
 				APPLICATION_EXTENSION_API_ONLY = YES;
 				CLANG_ENABLE_MODULES = YES;
-				CODE_SIGN_IDENTITY = "";
+				CODE_SIGN_IDENTITY = "iPhone Developer";
```
prasanthu: How did that happen ? Is this OK

prasanthu: It didn't break anything for me...

designatednerd: I believe it was changed because you added a framework that has to be signed to the debug target. I'm not 100% sure but basically I'm going with "if it didn't break anything, let it be."

PR #144, opened 2019-04-16
Changes after `bundle exec pod install`

prasanthu:


designatednerd:
@prasanthu I think #143 should address this since it also runs `pod update` (and changes some pods while it's at it)

designatednerd:
@prasanthu now that this is down to 0 changes, should we close this PR? 😄

prasanthu:
No changes...

PR #148, opened 2019-04-17
SwiftLint!

designatednerd:
In this PR: 
- Add SwiftLint and set it up to run on the Dev target (which should almost always be what we're building with locally)
- Fix almost 300 warnings from SwiftLint 🙀
- Remove a couple unused classes
- Add an extension on `UIApplication` which allows us to centralize accessing a properly typed delegate.
- Pulled a bunch of Netverify stuff out of two VC's into an extension
- Added some fun toys for generating strings to add to an array if a given variable is not nil


ostelco-ios-client/ViewControllers/SignUp/TheLegalStuffViewController.swift
```diff
@@ -15,25 +15,25 @@ class TheLegalStuffViewController: UIViewController {
         case termsAndConditions = "https://pi-redirector.firebaseapp.com/terms-and-conditions"
     }
     
-    @IBOutlet weak var termsAndConditionsLabel: UILabel!
-    @IBOutlet weak var privacyPolicyLabel: UILabel!
-    @IBOutlet weak var oyaUpdatesLabel: UILabel!
+    @IBOutlet private weak var termsAndConditionsLabel: UILabel!
```
prasanthu: Is this a usual practice to add `private` to @IBoutlet ?

designatednerd: yes - otherwise outside classes can re-set the label, which you don't want. 

For things like cells it's more common to use `private(set)`, so you can set the `text` on a label, but not the the label itself.

PR #154, opened 2019-04-24
Get purchase history from the backend

prasanthu:


ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+            }
+            self.records = purchases
+            self.tableView.reloadData()
+        }
+    }
+
+    // MARK: - Table view data source
+
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        // #warning Incomplete implementation, return the number of rows
```
designatednerd: You should probably delete this comment as you are returning the number of rows

ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+                        let sortedPurchases = purchases.sorted { $0.timestamp > $1.timestamp }
+                        let records: [PurchaseRecord] = sortedPurchases.map {
+                            let date = Date(timeIntervalSince1970: (Double($0.timestamp) / 1000.0))
+                            let strDate = dateFormatter.string(from: date)
+                            return PurchaseRecord(
+                                name: $0.product.presentation.label,
+                                amount: $0.product.presentation.price,
+                                date: strDate
+                            )
+                        }
```
designatednerd: Might be worth pulling `PurchaseRecord` out into its own file and making an initializer it can use from a `PurchaseModel` just to not have all this crap in the VC (and also be able to test it)

ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: "PurchaseItem", for: indexPath)
+        if let purchaseCell: PurchaseHistoryTableViewCell = cell as? PurchaseHistoryTableViewCell {
+            let record = self.records[indexPath.row]
+            purchaseCell.name.text = record.name
+            purchaseCell.date.text = record.date
+            purchaseCell.price.text = record.amount
+            return purchaseCell
+        } else {
+            return cell
```
designatednerd: For now this is fine, but remind me to walk you through some of the table view data source and generic cell dequeuing stuff I've set up in `core` to try to make this stuff a little easier

ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    func getPurchases(completionHandler: @escaping ([PurchaseRecord], Error?) -> Void) {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateStyle = .medium
+        dateFormatter.timeStyle = .none
```
designatednerd: I would recommend making this a `private lazy var` on the vc - DateFormatters are stupidly expensive to create, so only creating one for the VC is probably a good idea.

ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        refreshControl.addTarget(self, action: #selector(didPullToRefresh), for: .valueChanged)
+        tableView.refreshControl = refreshControl
+        didPullToRefresh()
+    }
+
+    @objc func didPullToRefresh() {
+        getPurchases { purchases, error in
+            self.tableView.refreshControl?.endRefreshing()
+            if let error = error {
+                print("error fetching purchases \(error)")
```
designatednerd: a) can you please use `debugPrint` for anything you intend to be a debugging statement?
b) Is there anything we can/should show users if this fails? 
c) Do we want to update the records and reload the data if an error occurs, or only if one doesn't occur?

ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
prasanthu: done

ostelco-ios-client/ViewControllers/Home/PurchaseHistoryTableViewController.swift
prasanthu: Ok

PR #155, opened 2019-04-25
Use the correct denomination while doing apple pay.

prasanthu:
Prime expects amount in currency’s smallest unit. For example, to
charge $10 USD, use an amount value of 1000 (i.e, 1000 cents).

### DIFF
--- a/ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
+++ b/ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
@@ -9,6 +9,7 @@
 import Stripe
 import Siesta
 
+
 extension UIViewController {
     func startApplePay(product: Product, delegate: PKPaymentAuthorizationViewControllerDelegate) {
         let merchantIdentifier = Environment().configuration(.AppleMerchantId)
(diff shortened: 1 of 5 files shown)


ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
```diff
@@ -61,4 +63,21 @@ extension UIViewController {
             #endif
         }
     }
+
+    // Stripe (& Prime) expects amounts to be provided in currency's smallest unit.
+    // https://stripe.com/docs/currencies#zero-decimal
+    // https://github.com/stripe/stripe-ios/blob/v15.0.1/Stripe/NSDecimalNumber%2BStripe_Currency.m
+    func convertStripeToNormalCurrency(amount: Decimal, currency: String) -> NSDecimalNumber {
```
designatednerd: is there a reason we can't just `import Stripe` here and use their implementation instead of copying it?

ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
```diff
@@ -9,6 +9,7 @@
 import Stripe
 import Siesta
 
+
```
designatednerd: SwiftLint should be yelling at you about this but: Excess newline

ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
prasanthu: I wasn't able to import that extension. Can you try?

ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
prasanthu: Done.

ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
designatednerd: yep, hold plz....

designatednerd: wait a minute, I'm gonna DM you, something's not quite adding up here

designatednerd: OK it's not just Stripe + Prime, it's `PKPaymentSummaryItem` as well. Whee!

PR #156, opened 2019-04-25
Country + EKYC with UI components

designatednerd:
This got *slightly* away from me as I was testing that everything was working, because I needed to get some stuff playing a bit nicer with location in order to do that. Fun times. 

In this PR: 
- Went through the EKYC and Country storyboards and updated everything to use appropriate labels and buttons
- Added a bunch of stuff around how we fetch locations. The primary thing is now `LocationController`, which lives in `ostelco-core`. There's also a `LocationChecking` protocol constrained to view controllers to create some opt-in functionality for being able to check location in various places
- Moved `Country` to `core`
- Updated a couple things to use single resolution PDFs instead of multi resolution pngs
- Added a couple illustrations
- Ripped out the multiple different view controllers around location problems and centralized them to a single one, using a `LocationProblem` enum that lives in `core` and an extension which lives in the main app (mostly to facilitate grabbing things from the bundle). 
- Updated SingPass/ID selection logic to match design

To come in future PRs (at least for these 2 storyboards): 
- Add new ID "Are you sure you want to do all these steps?" screen
- Centralize the stuff around address fill in / edit
- Add custom label to match iOS style header
- Update some OH NO! stuff (I'm probably going to pull this out into a separate storyboard)


ostelco-ios-client/Assets/Assets.xcassets/icIcon.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
...
+    {
+      "idiom" : "universal",
+      "filename" : "idIcon.pdf"
+    }
+  ],
+  "info" : {
+    "version" : 1,
+    "author" : "xcode"
+  }
+}
```
prasanthu: no new line at EOF

prasanthu: Is this something automatically created ?

ostelco-core/Models/Country.swift
```diff
@@ -7,28 +7,27 @@
...
+    public init(_ countryCode: String) {
         self.countryCode = countryCode
     }
-}
 
-extension Country: Equatable {
+    // MARK: - Equatable
     
-    static func == (lhs: Country, rhs: Country) -> Bool {
+    public static func == (lhs: Country, rhs: Country) -> Bool {
```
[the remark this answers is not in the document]
designatednerd: in theory that shouldn't be necessary but in practice that's probably a good idea

ostelco-ios-client/Assets/Assets.xcassets/icIcon.imageset/Contents.json
designatednerd: yes, this is auto-created by xcode, so I can't add the newline to it unfortunately. That red thing makes me bonkers as well

PR #157, opened 2019-04-25
Update Sign Up + ESim storyboards with styles, add Oh No consolidated storyboard

designatednerd:
NOTE: I haven't had an opportunity  to test the ESim stuff since I'm not presently able to get past identity verification. 

In this PR: 
- Updated UI in Sign Up and ESim storyboards to match updated designs
- Updated legal stuff to use checkmarks instead of switches and remove superflous legalese
- Created `OhNo` storyboard and a more flexible view controller to allow use of it from multiple places
- Made it easier to bold parts of a `BodyTextLabel`


ostelco-ios-client/ViewControllers/Login/LoginViewController.swift
```diff
@@ -108,7 +107,12 @@ class LoginViewController2: UIViewController {
     }
     
     @objc private func showEKYCOhNo() {
-        performSegue(withIdentifier: "showEKYCOhNo", sender: nil)
+        let ohNo = OhNoViewController.fromStoryboard(type: .ekycRejected)
+        ohNo.primaryButtonAction = {
+            // TODO: What are we actually retrying when this fails?
```
havard024: Retry means start ekyc process again, in practical terms it means send user to ekyc on boarding screen

designatednerd: Thanks for the reminder - I meant to update this when I saw it in Slack and then i got distracted 😛

PR #159, opened 2019-04-25
send user to ekyc oh no screen on rejected and check verification status when app will enter foreground

havard024:
Need to change EKYC storyboard, waiting for #157 to be merged


ostelco-ios-client/ViewControllers/EKYC/PendingVerificationViewController.swift
```diff
@@ -19,25 +19,68 @@ class PendingVerificationViewController: UIViewController {
...
-                    } else {
-                        if let status = regionResponse.kycStatusMap.JUMIO {
-                            self.showAlert(title: "Status", msg: status.rawValue)
-                        } else {
+                    case .PENDING:
+                        if !silentCheck {
                             self.showAlert(title: "Status", msg: regionResponse.status.rawValue)
                         }
+                    case .REJECTED:
+                        if let jumioStatus = regionResponse.kycStatusMap.JUMIO, let nricStatus = regionResponse.kycStatusMap.NRIC_FIN, let addressStatus = regionResponse.kycStatusMap.ADDRESS_AND_PHONE_NUMBER {
```
designatednerd: worth pulling this handling out into a separate method since it's got its own nested switch statement?

havard024: Yeah, could do that for each case on top level at least

designatednerd: That might even be overkill for the simpler stuff, but this one starts getting into quite a bit more logic, so it seems a bit more natural to pull out

PR #160, opened 2019-04-26
fix rejected case to be triggered if any is rejected instead of all

havard024:
also handle pending case by checking kycStatusMap since region status will be pending even if the jumio flow has failed

### DIFF
--- a/ostelco-ios-client/ViewControllers/EKYC/PendingVerificationViewController.swift
+++ b/ostelco-ios-client/ViewControllers/EKYC/PendingVerificationViewController.swift
@@ -52,10 +52,8 @@ class PendingVerificationViewController: UIViewController {
                     switch regionResponse.status {
                     case .APPROVED:
                         self.handleRegionApproved()
-                    case .PENDING:
-                        self.handleRegionPending(silentCheck: silentCheck)
-                    case .REJECTED:
-                        self.handleRegionRejected(silentCheck: silentCheck, regionResponse: regionResponse)
+                    default:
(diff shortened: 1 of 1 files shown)


ostelco-ios-client/ViewControllers/EKYC/PendingVerificationViewController.swift
```diff
@@ -81,14 +81,16 @@ class PendingVerificationViewController: UIViewController {
     func handleRegionPendingOrRejected(silentCheck: Bool = false, regionResponse: RegionResponse) {
         if let jumioStatus = regionResponse.kycStatusMap.JUMIO, let nricStatus = regionResponse.kycStatusMap.NRIC_FIN, let addressStatus = regionResponse.kycStatusMap.ADDRESS_AND_PHONE_NUMBER {
             switch (jumioStatus, nricStatus, addressStatus) {
-            case (let jumioStatus, let nricStatus, let addressStatus) where jumioStatus == .REJECTED || nricStatus == .REJECTED || addressStatus == .REJECTED:
+            // case (let jumioStatus, let nricStatus, let addressStatus) where jumioStatus == .REJECTED || nricStatus == .REJECTED || addressStatus == .REJECTED:
```
designatednerd: Perhaps this comment might be clearer if you just delete up to `where` and only show the or'd logic (same below)

havard024: sorry, meant to remove those comments!

PR #162, opened 2019-04-26
Refactor Apple Pay methods

prasanthu:
1) Refactoring the Apple Pay related methods
2) Move the formatting logic for Data balance label in home screen to DataAmountOnHomeLabel
3) Change Apple Pay button style


ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
```diff
@@ -9,25 +9,80 @@
...
-            return
+extension ApplePayError: RawRepresentable {
+    typealias RawValue = String
+
+    // Dont allow to construct using string
+    public init?(rawValue: RawValue) {
+        return nil;
+    }
+
+    public var rawValue: RawValue {
```
designatednerd: I would actually recommend just making `ApplePayError` conform to `LocalizedError`, and having these be `localizedDescription` (we can actually localize them later) - `rawValue` should not take on a different meaning here than it does in other places

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+            guard let source = source, error == nil else {
+                debugPrint(error!)
+                completion(PKPaymentAuthorizationResult(status: .failure, errors: [error!]))
+                self.showAlert(title: "Failed to create stripe source", msg: "\(error!.localizedDescription)")
+                return
+            }
+            // Call Prime API to buy the product.
+            APIManager.sharedInstance.products.child(product.sku).child("purchase").withParam("sourceId", source.stripeID).request(.post)
+                .onProgress({ progress in
+                    debugPrint("Progress %{public}@", "\(progress)")
```
designatednerd: You can do this just as `"Progress: \(progress)"` with `debugPrint` since it's not being logged when not actively debugging the way `os_log` can be

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+//  Created by Prasanth Ullattil on 26/04/2019.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import UIKit
+import Stripe
+import Siesta
+
+// Default implementaion for the PKPaymentAuthorizationViewControllerDelegate methods.
+class ApplePayViewController: UIViewController, ApplePayDelegate {
```
designatednerd: Is it possible to make this a protocol restricted to `UIViewController`? You can see an example of this with [`LocationCheckingViewController`](https://github.com/ostelco/ostelco-ios-client/blob/develop/ostelco-ios-client/Protocols/LocationCheckingViewController.swift), but it essentially allows us to have opt-in functionality without having to subclass.

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    func paymentAuthorizationViewControllerDidFinish(_ controller: PKPaymentAuthorizationViewController) {
+        // Dismiss payment authorization view controller
+        dismiss(animated: true, completion: {
+            if let applePayError = self.applePayError {
+                self.paymentError(applePayError)
+            } else if self.authorizedApplePay {
+                self.paymentSuccessful(self.purchasingProduct!)
```
designatednerd: why the force unwrap here?

OstelcoStyles/OstelcoLabel.swift
```diff
@@ -52,13 +52,75 @@ open class OstelcoLabel: UILabel {
...
+            let attributes: [NSAttributedString.Key: Any] = [
+                .font: smallFont,
+                .baselineOffset: offset,
+                .foregroundColor: color
+            ]
+            attrString.append(NSMutableAttributedString(string: decimalPart, attributes: attributes))
+        }
+        // Add the modifier part with bigger font.
+        attrString.append(NSMutableAttributedString(string: unit, attributes: [.font: bigFont, .foregroundColor: color]))
+        return attrString
```
designatednerd: This is fantastic - you can delete the `DataDecimalsLabel` below since this now covers both the normal amount and the decimals.

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
prasanthu: I tried, but was hitting the problem listed here. https://stackoverflow.com/questions/39487168/non-objc-method-does-not-satisfy-optional-requirement-of-objc-protocol

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
prasanthu: done

ostelco-ios-client/Extensions/UIViewController+StartApplePay.swift
prasanthu: done

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
prasanthu: done

OstelcoStyles/OstelcoLabel.swift
prasanthu: done

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
designatednerd: Ah yeah, so you can't just conform to a protocol with a protocol extension.  

There's a couple different workarounds to this but what I normally do is still implement the UIKit protocol on the view controller but just have it call through to a protocol extension I've defined. For example: 

```swift
/// MyProtocol.swift

// Protocol declaration
protocol MyProtocol {
   func handleAppleDidSomething() 
} 

// Default implementation of protocol
extension MyProtocol {
   func handleAppleDidSomething() {
      // Common handling
   }
}

/// MyViewController.swift

// Declaration of class
class MyViewController: UIViewController {
   // View controllery things
}

// Mix-in extension since we just want the behavior of the default implementation
extension MyViewController: MyProtocol { }

// Extension conforming to apple protocol that can't deal with @nonobjc things
extension MyViewController: AppleProtocol {
     func appleDidSomething() {
       self.handleAppleDidSomething()
    }
}
```

You also could constrain your default extension to only apply if the conforming object also implements `AppleProtocol`: 

```swift
extension MyProtocol where Self: AppleProtocol {
```

That way if the thing conforming to `MyProtocol` view controller doesn't also implement `AppleProtocol`, it doesn't get the default implementation for free.

ostelco-ios-client/ViewControllers/Home/HomeViewController2.swift
```diff
@@ -10,13 +10,14 @@ import UIKit
 import Stripe
 import Siesta
 import ostelco_core
+import OstelcoStyles
+
+class HomeViewController2: ApplePayViewController {
```
designatednerd: while you're mucking around in here you wanna drop the `2` off this? 😄

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
prasanthu: Thanks for the help. I have still retained the `ApplePayViewController` base class. That implements the `ApplePayDelegate` properties and `PKPaymentAuthorizationViewControllerDelegate` methods

ostelco-ios-client/ViewControllers/Home/HomeViewController2.swift
prasanthu: done

PR #163, opened 2019-04-28
Version Bump to 41

havard024:


OstelcoStyles/Info.plist
```diff
@@ -17,6 +17,6 @@
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
 	<key>CFBundleVersion</key>
-	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<string>41</string>
```
havard024: @designatednerd  We use increment_build_number in fastlane to bump the version number, but seems it's overwriting some other logic you have added

designatednerd: This is fine, we want that to get updated by `increment_build_number`

havard024: great thanks!

PR #164, opened 2019-04-29
add analytics helper

havard024:


ostelco-ios-client/Network/Analytics.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+//  Created by mac on 4/29/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import Foundation
+import ostelco_core
+import FirebaseAnalytics
+
+class ROAnalytics {
```
havard024: Firebase calls their analytics class for Analytics, so decided to prefix with our internal name RO (RedOtter)

designatednerd: Yeah that drives me batty, they really need to prefix that shit with `FIR`. 

FWIW I've been prefixing with `Ostelco` for anything that's within what we're building to disambiguate - `RO` doesn't really have a lot of meaning outside this app.

ostelco-ios-client.xcodeproj/project.pbxproj
```diff
@@ -1797,6 +1802,7 @@
 				042928F9223D4D8800806438 /* SignUpCompletedViewController.swift in Sources */,
 				04834D732237B11800A01500 /* CustomerModel.swift in Sources */,
 				C29802AD2181EDA800FBC5C3 /* OstelcoAPI.swift in Sources */,
+				04D6C9862276E32700AE13DD /* Analytics.swift in Sources */,
```
designatednerd: If you're gonna prefix a class, please also prefix the filename (makes it WAY easier to find)

ostelco-ios-client/Network/Analytics.swift
```diff
@@ -0,0 +1,60 @@
...
+
+extension AnalyticsEvent {
+    var name: String {
+        switch self {
+        case .ChosenCountry:
+            return "chosen_country"
+        case .ChosenIDMethod:
+            return "chosen_id_method"
+        default:
+            return String(describing: self).snakeCased()! // TODO: Is this safe enough to force unwrap?
```
designatednerd: I would just use `...snakeCased() ?? "" - I really try to avoid force unwrapping at all in anything related to analytics because crashing in Analytics calls is a pretty major 🤦‍♀️

ostelco-ios-client/Network/Analytics.swift
```diff
@@ -0,0 +1,60 @@
...
+            return ["country": country.countryCode]
+        case .ChosenIDMethod(let idMethod):
+            return ["id_method": idMethod]
+        default:
+            return [:]
+        }
+    }
+}
+
+// Found it on the internet: https://gist.github.com/dmsl1805/ad9a14b127d0409cf9621dc13d237457
```
designatednerd: You also could [steal `JSONEncoder`'s implementation](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/JSONEncoder.swift#L135) from the stdlib

ostelco-ios-client/Network/Analytics.swift
havard024: I agree, that sucks for users, if app crashes because of analytics 😓 
I'll add optional and a check before sending the event, if the string is empty, maby even send a warning to crashlytics

ostelco-ios-client.xcodeproj/project.pbxproj
havard024: Of course!

ostelco-ios-client/Network/Analytics.swift
havard024: @designatednerd By stealing, you mean copy / pasting that version inside the string extension?

designatednerd: Basically. I would double check the licenses we're using vs. they're using before doing that vs. re-implementing something similar, but my recollection is the stdlib license is quite permissive.

ostelco-ios-client/Network/OstelcoAnalytics.swift
```diff
@@ -0,0 +1,70 @@
...
+        var localizedDescription: String {
+            switch self {
+            case .eventNameIsEmpty:
+                return "Event name in analytics event is empty"
+            }
+        }
+    }
+    
+    static func logEvent(_ event: AnalyticsEvent) {
+
```
designatednerd: stray newline snuck in here

ostelco-ios-client/ViewControllers/ESim/ESIMOnBoardingViewController.swift
```diff
@@ -11,10 +11,12 @@ import UIKit
...
     @IBAction private func continueTapped(_ sender: Any) {
+        OstelcoAnalytics.logEvent(.DownloadingESIM)
         performSegue(withIdentifier: "showESIMInstructions", sender: self)
     }
     
     @IBAction private func needHelpTapped(_ sender: Any) {
         showNeedHelpActionSheet()
     }
 }
+
```
designatednerd: ^stray newline

ostelco-ios-client/Extensions/String+SnakeCase.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Created by mac on 4/29/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import Foundation
+
+// Found it on the internet: https://gist.github.com/dmsl1805/ad9a14b127d0409cf9621dc13d237457
+extension String {
+    
+    func snakeCased() -> String? {
```
designatednerd: 👍 good call pulling this out to an extension

ostelco-ios-client/Network/Analytics.swift
designatednerd: This has been pulled into an extension so we can deal with that as we see fit later.

PR #165, opened 2019-04-29
Fix the crash while calling the payment handlers.

prasanthu:
Read all the comments before implementing new design. The way I implemented will make it crash ( call the the methods `paymentAuthorizationViewController` recursively )


ostelco-ios-client/Protocols/ApplePayDelegate.swift
```diff
@@ -38,6 +38,7 @@ extension ApplePayError: LocalizedError {
     }
 }
 
+// swiftlint:disable:next class_delegate_protocol
```
designatednerd: Blergh, that's probably a swift 5 bug on swiftLint - now that you can specify that anything can be a protocol constraint (as UIViewController is here), SwiftLint keeps looking for the `ApplePayDelegate: class` declaration even though it's no longer necessary.

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -24,14 +24,14 @@ class ApplePayViewController: UIViewController {
     var applePayError: ApplePayError?
 
 }
-extension ApplePayViewController: ApplePayDelegate {
+extension ApplePayViewController: ApplePayDelegate, PKPaymentAuthorizationViewControllerDelegate {
```
designatednerd: fwiw i'd break this out into separate extensions - that way you can more clearly tell which are the apple methods you're overriding

prasanthu: Since I am only using default implementation for

prasanthu: extension ApplePayViewController: ApplePayDelegate is empty,

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -16,15 +16,16 @@ import PassKit
 // TL;DR: @objc functions may not currently be in protocol extensions.
 // You could create a base class instead, though that's not an ideal solution.
 // https://stackoverflow.com/questions/39487168/non-objc-method-does-not-satisfy-optional-requirement-of-objc-protocol
-class ApplePayViewController: UIViewController {
+class ApplePayViewController: UIViewController, ApplePayDelegate {
```
prasanthu: Is this a better way ? @designatednerd

PR #166, opened 2019-04-29
Allow methods defined in protocol to be re-implemented

prasanthu:


ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -18,11 +18,28 @@ import PassKit
 // https://stackoverflow.com/questions/39487168/non-objc-method-does-not-satisfy-optional-requirement-of-objc-protocol
 class ApplePayViewController: UIViewController, ApplePayDelegate {
     // MARK: - Properties for ApplePayDelegate.
+
     var shownApplePay = false
     var authorizedApplePay = false
     var purchasingProduct: Product?
     var applePayError: ApplePayError?
 
+    // MARK: - Default implementaion of ApplePayDelegate.
```
designatednerd: I would recommend removing this since this is now in a view controller superclass rather than in the default implementation of a protocol

PR #168, opened 2019-04-29
Make Become-A-Member-View be based on actual plans

prasanthu:


ostelco-ios-client/ViewControllers/Home/BecomeAMemberViewController.swift
```diff
@@ -15,54 +15,81 @@ class BecomeAMemberViewController: ApplePayViewController {
...
-            currency: "SGD",
-            sku: "123",
-            type: "plan"
-        )
-        startApplePay(product: product)
+        if let plan = plan {
+            startApplePay(product: plan)
+        } else {
+            debugPrint("No subscription plan found.")
+            self.showAlert(title: "Subscription Error", msg: "Did not find a valid subscription plan.")
```
designatednerd: This seems like a big enough error that we should probably be logging it to crashlytics

ostelco-ios-client/ViewControllers/Home/BecomeAMemberViewController.swift
```diff
@@ -15,54 +15,81 @@ class BecomeAMemberViewController: ApplePayViewController {
...
+        case .noSupportedCards?,
+             .otherRestrictions?:
+            showSetupButton = true
+        default:
+            showSetupButton = false
+        }
+        // Properties to Setup Apple Pay.
+        var action = #selector(BecomeAMemberViewController.setUpButtonTapped)
+        var paymentButtonType: PKPaymentButtonType =  .setUp
+        let paymentButton: PKPaymentButton
```
designatednerd: this can be moved below the if statement since it's not actually used in the if statement

ostelco-ios-client/ViewControllers/Home/BecomeAMemberViewController.swift
```diff
@@ -15,54 +15,81 @@ class BecomeAMemberViewController: ApplePayViewController {
...
+        if showSetupButton == false {
+            action = #selector(BecomeAMemberViewController.buyButtonTapped)
+            paymentButtonType =  .checkout
+        }
```
designatednerd: this is a little hard to reason about, particularly the `if showSetupButton == false` block.

You can set up the `action` and `paymentButtonType` as `let`s above the `if`/`else` statement (since they never change once assigned) and then assign them within the `if`/`else` statement before using them below. That way they can still be non-optional but it's a lot easier to parse visually when each case is happening.

ostelco-ios-client/ViewControllers/Home/BecomeAMemberViewController.swift
```diff
@@ -15,54 +15,81 @@ class BecomeAMemberViewController: ApplePayViewController {
...
+            return firstPlan
         }
-        
+        return nil
     }
-    
+
+    func setupPaymentButton() {
+        var showSetupButton = false
+        let applePayError:ApplePayError? = canMakePayments()
```
designatednerd: You should be seeing a SwiftLint warning on this line because of the colon spacing - please let me know if you're not.

ostelco-ios-client/ViewControllers/Home/BecomeAMemberViewController.swift
```diff
@@ -15,54 +15,81 @@ class BecomeAMemberViewController: ApplePayViewController {
...
+        switch applePayError {
+        case .unsupportedDevice?:
+            debugPrint("Apple Pay is not supported on this device")
```
designatednerd: we should probably have some kind of UI for this - mind throwing in a // TODO?

ostelco-ios-client/Network/AppErrors.swift
```diff
@@ -0,0 +1,26 @@
...
+//  ostelco-ios-client
+//
+//  Created by Prasanth Ullattil on 30/04/2019.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import Foundation
+import Crashlytics
+
+class ApplicationErrors {
```
prasanthu: @designatednerd @havard024 Is this a good way to report non fatal errors to crashalytics?

designatednerd: Yeah I think so, adding a wrapper like this will be helpful so if we ever need to change the underlying error reporting method it's just in one place.

havard024: Agreed, this seems fine, verified that they are visible in crashlytics for firebase

PR #171, opened 2019-04-30
Text Field Styling + Need Help? Re-adding

designatednerd:
In this PR: 

- Added custom text field styles and used them in the text edit cell and on the get started VC
- Nuked the no-longer-used "Main" storyboard and related class
- Added Need Help? buttons to the vc's in the Sign Up storyboard
- Poked all storyboards (except settings and home, which I think Prasanth is working in) to be using the iPhone Xs as their default size (smallest phone which supports esim)


OstelcoStyles/OstelcoTextField.swift
```diff
@@ -0,0 +1,150 @@
...
+    open override var placeholder: String? {
+        didSet {
+            guard let holder = self.placeholder else {
+                self.attributedPlaceholder = nil
+                return
+            }
+            
+            self.attributedPlaceholder = NSAttributedString(string: holder, attributes: [
+                .font: self.appFont.toUIFont,
+                .foregroundColor: self.appPlaceholderColor.toUIColor
```
designatednerd: This is the only way to change the font or color of placeholder text on a `UITextField`. Yes, this is silly. Nope, if they haven't changed it yet, they're not gonna.

ostelco-ios-client/ViewControllers/SignUp/GetStartedViewController.swift
```diff
@@ -21,7 +21,10 @@ class GetStartedViewController: UIViewController {
         
         self.hideKeyboardWhenTappedAround()
         self.continueButton.isEnabled = false
-        self.nameTextField.delegate = self
```
designatednerd: This is now set up in the storyboard

OstelcoStyles/OstelcoTextField.swift
havard024: interesting!

PR #172, opened 2019-05-01
Fix for empty eSimActivationCode

prasanthu:
A recent change in Prime has made the eSimActivationCode a nullable field.
This is a temporary fix for the crash we are getting while parsing the context


ostelco-ios-client/ViewControllers/ESim/ESIMPendingDownloadViewController.swift
```diff
@@ -55,7 +55,7 @@ class ESIMPendingDownloadViewController: UIViewController {
             .onSuccess { data in
                 if let simProfiles: [SimProfile] = data.typedContent(ifNone: nil) {
                     if let simProfile = simProfiles.first(where: {
-                        $0.eSimActivationCode == self.simProfile?.eSimActivationCode
+                        $0.iccId == self.simProfile?.iccId
```
designatednerd: might be worth throwing a `#warning` in here to fix this - this is definitely something we should fix before shipping

PR #174, opened 2019-05-01
Add and test header generator

designatednerd:
In anticipation of removing the current Siesta stuff, here's a basic header generator we can use a bit more flexibly than what we've presently got.

### DIFF
--- a/ostelco-core/Network/Headers.swift
+++ b/ostelco-core/Network/Headers.swift
@@ -25,7 +25,7 @@ public enum HeaderValue {
         case .applicationJSON:
             return "application/json"
         case .token(let token):
-            return token
+            return "Bearer \(token)"
         }
     }
 }
--- a/ostelco-core/Network/Headers.swift
+++ b/ostelco-core/Network/Headers.swift
@@ -1,0 +1,88 @@
+//
+//  Headers.swift
+//  ostelco-core
+//
+//  Created by Ellen Shapiro on 5/1/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import Foundation
+
+/// Keys to include in headers. The raw value is the key for a header dict.
+public enum HeaderKey: String {
+    case authorization = "Authorization"
+    case contentType = "Content-Type"
+}
+
+/// Values to include in headers.
+/// `toString` is the value which should be included in a header dictionary.
+public enum HeaderValue {
+    case applicationJSON
+    case token(String)
+    
+    public var toString: String {
+        switch self {
+        case .applicationJSON:
+            return "application/json"
+        case .token(let token):
+            return token
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


ostelco-core/Network/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+public enum HeaderValue {
+    case applicationJSON
+    case token(String)
+    
+    public var toString: String {
+        switch self {
+        case .applicationJSON:
+            return "application/json"
+        case .token(let token):
+            return token
```
designatednerd: Do we need to prefix this with `Bearer ` for all our APIs, or just some of them? It wasn't totally clear looking at the existing ones.

havard024: all OstelcoAPIs required `Authorization: Bearer {{ JWT }}`

designatednerd: @havard024 Just the actual access token, right? Looking at [stuff like this](https://github.com/ostelco/ostelco-ios-client/blob/develop/ostelco-ios-client/Network/Auth.swift#L83) it seems like that's what's being set up at the moment.

havard024: @designatednerd Ah that stuff, so the auth header logic was implemented first, for OstelcoAPI, there we need to store the full header value with Bearer in front.
Then a new use case appeared inside [GetStartedViewController](https://github.com/ostelco/ostelco-ios-client/blob/6d68afdb6ca7b1bb664ea40fe9c2686a53983054/ostelco-ios-client/ViewControllers/SignUp/GetStartedViewController.swift#L58) we needed to access the email inside the JWT to be able to pass it to the server to create a new user (auth0 and our server is disconnected, so even if you are logged in, that doesn't mean you have a user in our systems).
That's why it's stored two places, I wasn't able to figure out how to store the raw value in Siesta, without the Bearer prefix, and I didn't think about getting the token from Siesta, parsing the header to extract the JWT, instead I stored the raw token some other place

designatednerd: OK so for newer things that are going to use this to construct headers though, we just need that API token that was being set up where I linked, correct? I can get probably figure out a different workaround later for the JWT thing (and this isn't actively being used yet, just built and tested first)

havard024: The token is also being set in the [SplashViewController](https://github.com/ostelco/ostelco-ios-client/blob/6297eb39b3f7bd22da1fe18dc1c5d81a4e1300e8/ostelco-ios-client/ViewControllers/SplashViewController.swift#L40)

designatednerd: Right - not replacing that yet. going to work on some consolidation stuff first

designatednerd:
Considering this slack an approval and merging
Screen Shot 2019-05-02 at 3 56 04 PM

PR #176, opened 2019-05-02
redirect user based on simProfile status

havard024:


ostelco-ios-client/Models/Region.swift
```diff
@@ -6,6 +6,14 @@
 //  Copyright © 2019 mac. All rights reserved.
 //
 
+func getSimProfileFromSimProfilesArray(_ simProfiles: [SimProfile]) -> SimProfile? {
```
havard024: Not much logic in there, but copied the way we've implemented fetching a single region from a list of regions in the [Context Model](https://github.com/ostelco/ostelco-ios-client/blob/6297eb39b3f7bd22da1fe18dc1c5d81a4e1300e8/ostelco-ios-client/Models/Context.swift#L11)

Might be more logic to select simProfile later, or maby not

ostelco-ios-client/Models/Region.swift
```diff
@@ -6,6 +6,14 @@
...
+    if simProfiles.isNotEmpty {
+        return simProfiles[0]
```
designatednerd: i would just do `return simProfiles.first` here - that returns `nil` if the array is empty

ostelco-ios-client/Storyboards/Login.storyboard
```diff
@@ -147,6 +148,14 @@
             </objects>
             <point key="canvasLocation" x="791" y="186"/>
         </scene>
+        <!--TabController-->
```
designatednerd: why does this need an additional reference to the tab controller here?

ostelco-ios-client/Models/Region.swift
designatednerd: Particularly if you're just grabbing the first one, I think this might be premature optimization. You can always add it later!

ostelco-ios-client/Storyboards/Login.storyboard
havard024: > This looks legit, but is there a particular reason not to wait until we consolidate this logic to one place?

yes, so we can make the awesome screen a part of the user testing next week

havard024: > why does this need an additional reference to the tab controller here?

this is the entry point to the home screen

ostelco-ios-client/Models/Region.swift
havard024: will do!

PR #179, opened 2019-05-03
Start moving some APIs to `LoggedInAPI` for realsies

designatednerd:
Apologies, this one slightly got away from me in terms of size. 

In this PR: 
- Add and test Request generator 
- Pull some stuff into a `BasicNetwork` superclass so we can pass arbitrary requests to it
- Pull a bunch of region response stuff into core
- Test picking a region out of a number of regions
- Consolidate error handling
- Add a temporary wrapper to let the `LoggedInAPI` use creds from Auth0
- Set up some stuff around API path generation to try to keep it type-safe
- Centralize looking for server errors in returned data
- Actually move a couple calls in the code over to new API!
- Nuke unused `OstelcoAPI`

### DIFF
--- a/ostelco-core/Network/APIEndpoint.swift
+++ b/ostelco-core/Network/APIEndpoint.swift
@@ -1,0 +1,64 @@
+//
+//  APIEndpoint.swift
+//  ostelco-core
+//
+//  Created by Ellen Shapiro on 5/1/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import Foundation
+
+protocol APIEndpoint {
+    var value: String { get }
+}
+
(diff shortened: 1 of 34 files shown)


ostelco-core/Network/APIEndpoint.swift
```diff
@@ -0,0 +1,71 @@
...
+    case region(code: String)
+    case profile
+    case simProfiles
+    case scans
+    
+    var value: String {
+        switch self {
+        case .dave:
+            // I can't let you do that,
+            return "dave"
```
designatednerd: Anyone know who Dave is or why his name is in the API request path? 😛

PR #180, opened 2019-05-03
Migrate more calls to LoggedInAPI

designatednerd:
In this PR: 
- Migrated most things using `load()` for `GET` API calls from Siesta to `LoggedInAPI`, which is using promise wrappers and handling data parsing under the hood.
- Removed no-longer-needed transformers from `APIManager`
- Moved `MyInfoDetails` to core. 
- Removed deprecated `Product` initializer and gave it a debugDescription

Future PRs (ie, things NOT done in this PR): 
- Update base pipeline based on final auth decision. **NOTE: This means auto-refreshing of tokens won't work on the migrated calls until that's done.** 
- Migrate the `context` calls using `load()` - right now that whole thing is a rat's nest in two different places, and I wanted to try and focus on some more low-hanging fruit first.
- Migrate stuff using `request` (most of which requires a bit more work).

### DIFF
--- a/ostelco-core/Models/MyInfoDetails.swift
+++ b/ostelco-core/Models/MyInfoDetails.swift
@@ -8,16 +8,32 @@
 
 import Foundation
 
-struct MyInfoAddress: Codable {
-    let country: String?
-    let unit: String?
-    let street: String?
-    let block: String?
-    let postal: String?
-    let floor: String?
-    let building: String?
+public struct MyInfoAddress: Codable {
+    public let country: String?
+    public let unit: String?
+    public let street: String?
+    public let block: String?
+    public let postal: String?
+    public let floor: String?
+    public let building: String?
     
-    func getAddressLine1() -> String {
+    public init(country: String?,
+                unit: String?,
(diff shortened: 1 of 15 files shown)


ostelco-core/Models/MyInfoDetails.swift
```diff
@@ -8,16 +8,32 @@
...
 
-struct MyInfoAddress: Codable {
-    let country: String?
-    let unit: String?
-    let street: String?
-    let block: String?
-    let postal: String?
-    let floor: String?
-    let building: String?
+public struct MyInfoAddress: Codable {
```
havard024: Any reason why we add public, is it good practice or required? (just out of curiosity)

designatednerd: Yep, it's because it's being moved to Core. Core is a separate framework from the main application, and for anything to be seen outside the framework, it must be public.

The good news is with a `public let` since it's only created at instantiation, outside callers can't change it. That is *not* the case for non-computed `public var`s, so be aware of that.

PR #181, opened 2019-05-05
Version Bump to 44

havard024:


ostelco-ios-client/Protocols/LocationCheckingViewController.swift
```diff
@@ -52,7 +52,7 @@ extension LocationChecking {
         
         self.spinnerView = showSpinner(onView: view, loadingText: "Checking location...")
         var isDebug = false
-        #if DEBUG
+        #if ENABLE_SKIP_USER_LOCATION_CHECK
```
havard024: Since we might want to deploy release builds where we can toggle flags on / off, I tried to replace the DEBUG flag with custom flags. Though I think it would be better if you have a env flag or config file where we set the value to true / false so we can clearly see which flags that are available and which flags is on / off

designatednerd: This works for now, but I agree, you'd want to have that behind some kind of settable bool (and also something that triple-checks we're not disabling this in prod)

PR #183, opened 2019-05-07
Feature/cancel membership

havard024:
Only one string needed change, else everything else is according to design


havard024:
You are right! I think I manually changed the file name, I should've done refactor through xcode, I'll fix it!

havard024:
@designatednerd now there are storyboard changes as well!

PR #185, opened 2019-05-07
replace alert with open external url in settings view controller

havard024:


havard024:
I believe it's only two places for now, settings and legal stuff screen

PR #186, opened 2019-05-07
Feature/remove payment todos

prasanthu:


ostelco-core/Models/ProductModel.swift
```diff
@@ -32,10 +32,10 @@ public struct ProductModel: Codable {
...
     public var type: String {
-        if let productType = properties["productType"] {
-            return productType.lowercased()
+        if let productClass = properties["productClass"] {
+            return productClass.lowercased()
         }
-        // default type is offer
-        return "offer"
+        // default type is simple_data
+        return "simple_data"
```
designatednerd: Is it worth using `Codable` for parsing the `Properties` into some kind of object where you can do this in a typed fashion instead of using strings? Basically something where `Properties` would have a `ProductClass` that would take a value from an enum? That way you could compare that rather than raw strings.

ostelco-ios-client/ViewControllers/Home/HomeViewController.swift
```diff
@@ -56,7 +56,7 @@ class HomeViewController: ApplePayViewController {
 
     private func checkForSubscription(_ products: [Product]) -> Bool {
         // See if the list contains offers.
-        let hasOffers = products.contains { $0.type == "offer" }
+        let hasOffers = products.contains { $0.type != "plan" && $0.currency.lowercased() == "sgd" }
```
designatednerd: Might be worth throwing in a `TODO` to get rid of the singapore dollars check when we take this international

ostelco-core/Models/ProductModel.swift
prasanthu: More changes are expected in the backend around this. It's not worth spending time until that is finalized.

designatednerd: 👍Fair enough

PR #188, opened 2019-05-09
Firebase Pt 1: Getting a working magic link login

designatednerd:
Sorry, this one's gonna be huge - this is basically everything it took to get the magic link authentication working and directing to proper places with the new firebase stuff

In this PR: 
- Swapped out authentication from Auth0 to firebase
- Updated to using magic link email validation from firebase
- Dev Art'd a couple of screens for entering your email and waiting for a magic link
- Added a `RootCoordinator` accessible from the app delegate which can be used to navigate the user around without getting into a nest of crazy segues
- Moved a bunch of stuff to the `UserManager` including parsing what screen the user needs to be sent to
- Added and tested a basic email validator
- Swapped out some stuff for providing the token to the `Header` and `Request` generators
- Removed keychain stuff because Firebase handles that for us

To come before merging to master: 
- Move the rest of the API calls to the new API
- Get rid of siesta and RxSwift and `FirebaseUI` stuff which it turns out we don't actually need with magic link authentication

To come after merging to master: 
- Test ALL THE THINGS!


ostelco-ios-client/Network/EmailLinkManager.swift
```diff
@@ -0,0 +1,84 @@
...
+                if let firebaseError = error {
+                    seal.reject(firebaseError)
+                    return
+                }
+                
+                guard authDataResult?.user != nil else {
+                    seal.reject(Error.noErrorAndNoUser)
+                    return
+                }
+
```
designatednerd: should nuke the pending email once login is successful

ostelco-ios-client/Network/UserManager.swift
```diff
@@ -7,35 +7,139 @@
 //
 
 import Crashlytics
+import FirebaseAuth
 import ostelco_core
+import PromiseKit
 
-class UserManager {
+class UserManager: NSObject {
```
designatednerd: this probably doesn't still need to be an `NSObject` subclass since it's not a firebase UI delegate anymore

ostelco-ios-client/ViewControllers/Email/EmailEntryViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+            self.continueButton.isEnabled = false
+            self.errorLabel.text = description
+        }
+    }
+}
+
+extension EmailEntryViewController: UITextFieldDelegate {
+    
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+
```
designatednerd: whitespace

designatednerd:
Prasanth is tied up so I'm gonna merge this one

PR #189, opened 2019-05-09
Firebase pt. 2: Refactor all APIs to use PrimeAPI

designatednerd:
In this PR: 
- Rename `LoggedInAPI` to `PrimeAPI` since all calls to prime have to be logged in 
- Migrate all remaining calls to `PrimeAPI`
- Update singleton naming to better match swift conventions
- Add a convenience PromiseKit wrapper for Stripe
- Delete all references to Siesta and RxSwift
- Add ability to login by pasting a link into the simulator (this is mostly for developer convenience - actually pasting the link into Safari was, for some reason, not working at all)

### DIFF
--- a/ostelco-core/Models/EKYCProfileUpdate.swift
+++ b/ostelco-core/Models/EKYCProfileUpdate.swift
@@ -1,0 +1,20 @@
+//
+//  EKYCProfileUpdate.swift
+//  ostelco-core
+//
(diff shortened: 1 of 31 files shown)


designatednerd:
Again, since Havard is out, Prasanth is tied up, and this passed CI, mergin'.

PR #190, opened 2019-05-09
Firebase Pt. 3: Remove unneeded pods

designatednerd:
In this PR: 
- Remove a whole mess of pods 
- Add a pod for a network mocking framework I'm going to be using later
- don't. touch. anything! else.


designatednerd:
Since this is just pods, I'm gonna merge it.

PR #195, opened 2019-05-10
Feature/fix login crash

prasanthu:


ostelco-ios-client/ModelControllers/UserDefaultsWrapper.swift
```diff
@@ -65,7 +65,11 @@ struct UserDefaultsWrapper {
             return self.value(for: .pendingEmail)
         }
         set {
-            self.setValue(newValue, for: .pendingEmail)
+            if let newEmail = newValue {
+                self.setValue(newEmail, for: .pendingEmail)
+            } else {
+                self.removeValue(for: .pendingEmail)
+            }
```
designatednerd: 🤦‍♀ Good catch, i'll update some tests to validate this

PR #196, opened 2019-05-10
Fix the "Create" APIs

prasanthu:
Most of our create APIs require parameters to be passed via Query
parameters.

### DIFF
--- a/ostelco-core/Network/PrimeAPI.swift
+++ b/ostelco-core/Network/PrimeAPI.swift
@@ -101,8 +101,9 @@ open class PrimeAPI: BasicNetwork {
         ]
         
         let path = RootEndpoint.products.pathByAddingEndpoints(productEndpoints)
-        
-        return self.sendObject(payment, to: path, method: .POST)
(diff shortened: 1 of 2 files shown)


ostelco-core/Network/PrimeAPI.swift
```diff
@@ -118,7 +119,11 @@ open class PrimeAPI: BasicNetwork {
     /// - Parameter userSetup: The `UserSetup` to use.
     /// - Returns: A promise which when fullfilled will contain the created customer model.
     public func createCustomer(with userSetup: UserSetup) -> Promise<CustomerModel> {
-        return self.sendObject(userSetup, to: RootEndpoint.customer.value, method: .POST)
+        let queryItems = [
+            URLQueryItem(name: "nickname", value: userSetup.nickname),
+            URLQueryItem(name: "contactEmail", value: userSetup.contactEmail)
```
designatednerd: Ugh, missed that `withParam` was query parameters even for `POST` requests. Sorry about that. 

...also why are we using query parameters instead of JSON for `POST` requests?

PR #199, opened 2019-05-13
Make sure that message labels are shown only once

prasanthu:


ostelco-ios-client/ViewControllers/Home/HomeViewController.swift
```diff
@@ -49,6 +49,7 @@ class HomeViewController: ApplePayViewController {
 
     private func showWelcomeMessage() {
         if HomeViewController.newSubscriber {
+            HomeViewController.newSubscriber = false
```
designatednerd: This is not persisted anywhere, so any time the user restarts the device, they'll still see the welcome message. Is that the behavior you're looking for?

designatednerd: Thank you for the clarification!

PR #200, opened 2019-05-13
A whole mess more tests and mocks!

designatednerd:
In this PR: 
- Added a whole bunch more tests and mocks - got a few things on the live tests now too. 
- Centralized providing mock data
- Got rid of unused API + data type


dev-ostelco-ios-clientTests/APITests/MockAPITests.swift
```diff
@@ -226,4 +330,48 @@ class MockAPITests: XCTestCase {
...
+        })
+        
+        // Failure handled in `awaitResult`
+        self.mockAPI.addAddress(address, forRegion: "sg").awaitResult(in: self)
+    }
+    
+    func testMockCreatingJumioScan() {
+        self.stubPath("regions/sg/kyc/jumio/scans", toLoad: "create_jumio_scan")
+        
+        guard let scan = self.mockAPI.createJumioScanForRegion(code: "sg").awaitResult(in: self) else {
```
prasanthu: I thought the fix for createJumioScanForRegion was in this PR. This is using `GET` instead of `POST`

prasanthu: Use this for reference: https://github.com/ostelco/ostelco-core/blob/develop/prime/infra/dev/prime-customer-api.yaml

designatednerd: @prasanthu It [is in this PR](https://github.com/ostelco/ostelco-ios-client/pull/200/files#diff-7a5823cfae457badcd2f8a8fb5087f4bR2130) and it's using `POST` now

PR #205, opened 2019-05-14
Feature/use myinfo config

prasanthu:


ostelco-ios-client/ViewControllers/EKYC/SelectIdentityVerificationMethodViewController.swift
```diff
@@ -60,23 +61,37 @@ class SelectIdentityVerificationMethodViewController: UIViewController {
...
+                self?.spinnerView = nil
+            }
+            .done { [weak self] myInfoConfig in
+                debugPrint("MyInfoConfig.url: \(myInfoConfig.url)")
+                var components = URLComponents(string: myInfoConfig.url)!
+                // Add purpose and state parameters to the url.
+                // state parameter is currently ignored by prime.
+                components.queryItems = [
+                    URLQueryItem(name: "purpose", value: "eKYC"),
+                    URLQueryItem(name: "state", value: "123")
```
designatednerd: Why do we need the state parameter if it's ignored?

ostelco-ios-client/ViewControllers/EKYC/SelectIdentityVerificationMethodViewController.swift
```diff
@@ -60,23 +61,37 @@ class SelectIdentityVerificationMethodViewController: UIViewController {
...
+                self?.showGenericError(error: error)
+            }
+
     }
     
-    func getMyInfoToken() {
-        if let url = getMyInfoURL() {
-            print("URL for API \(url.absoluteString)")
+    func showMyInfoLogin(url: URL?) {
+        if let url = url {
```
designatednerd: You may want to make this a `guard let` and record an error if the URL is nil here

ostelco-ios-client/ViewControllers/EKYC/SelectIdentityVerificationMethodViewController.swift
prasanthu: It is mandatory  for MyInfo, we could use this in our backend for verification.  The uniqueness of this parameter is not imposed by MyInfo 
https://www.ndi-api.gov.sg/library/trusted-data/myinfo/tutorial2

prasanthu: Prime is not using this now

designatednerd: OK - can you please update the comment to include the bit about it being mandatory for myinfo?

PR #207, opened 2019-05-14
Even more API tests!

designatednerd:
In this PR: 
- Added a whole mess more mock + live API tests (again, live API tests don't run on the server, and won't run locally unless you have a logged in user).
- Updated Stripe ephemeral token API to return a non optional `[String: AnyHashable]` (which is what auto-converts to `NSDictionary` nowadays)
- Moved some formatting stuff to `MyInfoAddress` and added tests around it
- Updated `Codable` things that are being sent as query params to have `asQueryParams` properties so that's all still under the hood of the object

### DIFF
--- a/ostelco-core/Network/PrimeAPI.swift
+++ b/ostelco-core/Network/PrimeAPI.swift
@@ -139,10 +139,9 @@ open class PrimeAPI: BasicNetwork {
     }
 
     /// - Returns: A Promise which when fulfilled will contain the Stripe Ephemeral Key
-    public func stripeEphemeralKey(stripeAPIVersion: String) -> Promise<[String: AnyObject]?> {
+    public func stripeEphemeralKey(with request: StripeEphemeralKeyRequest) -> Promise<[String: AnyObject]?> {
         let path = RootEndpoint.customer.pathByAddingEndpoints([CustomerEndpoint.stripeEphemeralKey])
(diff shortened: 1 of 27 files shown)


ostelco-core/Network/PrimeAPI.swift
```diff
@@ -144,11 +139,16 @@ open class PrimeAPI: BasicNetwork {
     }
 
     /// - Returns: A Promise which when fulfilled will contain the Stripe Ephemeral Key
-    public func stripeEphemeralKey(stripeAPIVersion: String) -> Promise<[String: AnyObject]?> {
+    public func stripeEphemeralKey(with request: StripeEphemeralKeyRequest) -> Promise<[String: AnyHashable]> {
```
prasanthu: Stripe expects the key as `[AnyHashable : Any]?`. hope this change will still work 
https://stripe.dev/stripe-ios/docs/Type%20Definitions.html#/c:STPBlocks.h@T@STPJSONResponseCompletionBlock

designatednerd: `String` conforms to `AnyHashable` and `AnyHashable` definitely conforms to `Any`. However, since it's clear that's the type they want, I'll update it. 

While you can pass in a nullable object because this is obj-c code, I can't imagine any scenario in which passing `nil` would lead to success there, which is why I'm having this throw if it doesn't get anything back.

PR #208, opened 2019-05-15
Tests + A few UI Things

designatednerd:
This one's a bit of a grab bag.

In this PR: 
- Add some tests of the Country Data Source
- Mock test fetching `MyInfo` from the backend
- Add UI to allow retrying when MyInfo fetch fails
- Update UI for happier autolayout in the my info details vc
- Get rid of xcasset colors and update storyboards to get rid of all named colors
- Remove theme, add simpler switch between dev/prod blue
- Hook up a bunch of stuff with IBOutlets so it can get tinted the correct color
- Get rid of now-unused extensions


OstelcoStyles/OstelcoLabel.swift
```diff
@@ -123,6 +123,16 @@ public class DataAmountOnHomeLabel: OstelcoLabel {
     }
 }
 
+public class DataRemainingLabel: OstelcoLabel {
```
prasanthu: Where is this used?  we already have `DataAmountOnHomeLabel`  to show the balance

ostelco-ios-client/AppDelegate.swift
```diff
@@ -65,6 +59,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
     
+    func configureAppearance() {
+        if
+            let bundleIndentifier = Bundle.main.bundleIdentifier,
```
prasanthu: This `if` code style looks weird

OstelcoStyles/OstelcoLabel.swift
designatednerd: It's the little "Left" label underneath that.

ostelco-ios-client/AppDelegate.swift
designatednerd: ah i need to update the indent on setting `useDevColor`, then it'll be a bit more reasonable. The idea is to put each thing we're checking on its own line and have them all line up. I admit it looks better with `guard` statements, but I've found it to be helpful to be able to parse multiple conditions as I'm reading.

designatednerd:
I have been testing the UI changes as I make them, i'm going through and hitting the APIs in the live app. It's actually part of why I started some of the changes, becuase the UI was looking a bit misaligned.

PR #211, opened 2019-05-15
Send more errors to Crashlytics

designatednerd:
Went through and audited what we're sending to Crashlytics and made sure that: 
- All errors coming through the API or other `catch` blocks in PromiseKit that are not explicitly handled get logged
- We now have an `ApplicationErrors.assertAndLog` method that allows you to have an assertion failure at the point where `assertAndLog` is called during dev, and to send an error to crashlytics in production

### DIFF
--- a/ostelco-ios-client/Network/AppErrors.swift
+++ b/ostelco-ios-client/Network/AppErrors.swift
@@ -13,11 +13,14 @@ import ostelco_core
 /// Wrapper for logging errors throughout the application
 struct ApplicationErrors {
     enum General: LocalizedError {
+        case assertionFailed(message: String, file: String, line: UInt)
         case noValidPlansFound
         case noMyInfoConfigFound
 
         var localizedDescription: String {
             switch self {
+            case .assertionFailed(let message, let file, let line):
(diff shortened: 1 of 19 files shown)


ostelco-ios-client/Network/AppErrors.swift
```diff
@@ -32,16 +35,54 @@ struct ApplicationErrors {
...
+    /// Logs a non-fatal error in production builds.
+    ///
+    /// - Parameters:
+    ///   - message: The message to include with the assertion failure
+    ///   - file: The file where this method was called. Defaults to the direct caller.
+    ///   - line: The line where this method was called. Defaults to the direct caller.
+    static func assertAndLog(_ message: String,
+                             file: StaticString = #file,
+                             line: UInt = #line) {
+        assertionFailure(message, file: file, line: line)
```
designatednerd: This is the secret sauce that allows the assertion failure to show up at the proper file and line in the code: By passing the file and line of the caller, that assertion failure will actually show up at the spot of the caller. Very similar to how this works with `XCTest` assertions.

PR #216, opened 2019-05-20
Pull Push Notification Handling Into a helper and test it

designatednerd:
In this PR: 
- Pulled all push notificaiton handling that we're doing in-app out of the app delegate and a few other places and centralized it in a `PushNotificationController` 
- Turned off the firebase app delegate proxy to allow that
- Turned on the ability of `Messaging` to receive messages without having to go through APNS while in the foreground. 
- Tested a bunch of stuff on the push controller that's within our control
- Updated the `UserManager` to be in `ModelControllers` and to hide the implementation detail of how we know a user is logged in (mostly for easier testing). 
- Added mix-in protocols for registering for local notifications for both push notifications and `applicationDidBecomeActive`. 

Still TK: 
- Handling actual push notifications coming from either Apple or from FCM directly


designatednerd:
There's been some gooberyness in the testing with it for some reason causing live data to be sent to the mock user helper, but as far as I can tell only on CI - I *think* I've got that nailed down so I'm going to merge this, but I'll revisit if it keeps being shitty.

PR #219, opened 2019-05-20
Feature/add payment errors

prasanthu:


ostelco-ios-client/Protocols/ApplePayDelegate.swift
```diff
@@ -139,6 +142,21 @@ extension ApplePayDelegate where Self: PKPaymentAuthorizationViewControllerDeleg
 
     // MARK: - Other Helpers.
 
+    // Create appropriate ApplePayError based on the Error
+    func createApplePayError(_ error: Error) -> ApplePayError {
+        // All generic payment errors are treated as primeAPIError
+        var paymentError = ApplePayError.primeAPIError(error)
+        if let apiError = error as? APIHelper.Error {
+            if case let .jsonError(jsonRequestError) = apiError {
+                // Payment is declined when status = 403 and code = FAILED_TO_PURCHASE_PRODUCT
```
prasanthu: Is there a better way to check this ?

designatednerd: I would probably `switch` on the error you pass in and return from that switch statement if it's a special case, otherwise return the default value: 

```swift
switch error {
case APIHelper.Error.jsonError(let jsonError): 
    // Handle status logic here, return payment declined if needed otherwise keep going
default: 
   break
}

return ApplePayError.primeAPIError(error)
```

ostelco-ios-client/ViewControllers/OhNoViewController.swift
```diff
@@ -54,13 +63,19 @@ class OhNoViewController: UIViewController {
                 return NSAttributedString(string: "Something went wrong.\n\nTry again in a while, or contact support")
             case .myInfoFailed:
                 return NSAttributedString(string: "We're unable to retrieve your info from MyInfo.\n\n. Try later.")
+            case .paymentFailedGeneric:
+                 return NSAttributedString(string: "Something went wrong in our system. We have not taken any money from your account. Try again in a while or contact customer support.")
```
designatednerd: i think the alignment is off here

ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -61,18 +61,9 @@ class ApplePayViewController: UIViewController, ApplePayDelegate {
 
     func showOhNo(type: OhNoViewController.IssueType) {
         let ohNo = OhNoViewController.fromStoryboard(type: type)
+        ohNo.buttonTitle = "Ok I understand"
```
designatednerd: I would recommend putting this in the `IssueType`'s property for the title. If it doesn't have a property for that yet, add one 😄

designatednerd:
Test failure is one of the occasional instances where a real token comes in to the mock token controller. I'm still not sure how in the hell that's possible, I'll look again in another branch, it doesn't need to be dealt with here.

PR #221, opened 2019-05-21
Push Notification Parsing

designatednerd:
In this PR: 
- Parsing push based on what we're getting from various push sources
- Fix (I THINK) issue with real push token getting set on Mock push controller.

Next PR: 
- Actual handling of data coming from Firebase

### DIFF
--- a/ostelco-ios-client/ModelControllers/PushNotificationController.swift
+++ b/ostelco-ios-client/ModelControllers/PushNotificationController.swift
@@ -35,9 +35,11 @@ class PushNotificationController: NSObject {
         super.init()
         
         UNUserNotificationCenter.current().delegate = self
-        Messaging.messaging().delegate = self
         
         if !UIApplication.isTesting {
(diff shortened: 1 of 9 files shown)


ostelco-ios-client/ModelControllers/PushNotificationController.swift
```diff
@@ -35,9 +35,11 @@ class PushNotificationController: NSObject {
         super.init()
         
         UNUserNotificationCenter.current().delegate = self
-        Messaging.messaging().delegate = self
         
         if !UIApplication.isTesting {
+            // Actually get messages sent through firebase's Messaging
+            Messaging.messaging().delegate = self
```
designatednerd: I think this is what's causing the firebase messaging token to get set as a real token on the mock. Not setting the delegate while testing should, in theory, mean that even if a delegate method is called, it won't go through to the mock notification controller.

PR #225, opened 2019-05-23
Add application version and build number

prasanthu:


ostelco-ios-client/ViewControllers/Home/SettingsViewController.swift
```diff
@@ -0,0 +1,25 @@
...
+class SettingsViewController: UIViewController {
+
+    @IBOutlet private weak var versionLabel: UILabel!
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        var appVersion = ""
+        if let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
+            appVersion = "App Version \(version)"
+        }
+        if let buildNumber = Bundle.main.infoDictionary?["CFBundleVersion"] as? String {
```
designatednerd: There's a `kCFBundleVersionKey` constant so you don't have to hard-code this string (stupidly, there's not one for the short version)

PR #226, opened 2019-05-23
Show topped up message

prasanthu:


ostelco-ios-client/ViewControllers/Home/HomeViewController.swift
```diff
@@ -47,18 +47,28 @@ class HomeViewController: ApplePayViewController {
         return formatter
     }()
 
+    private func showMessage() {
+        welcomeLabel.isHidden = false
+        messageLabel.isHidden = false
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) { [weak self] in
+            self?.welcomeLabel.isHidden = true
+            self?.messageLabel.isHidden = true
+        }
```
designatednerd: I can see this was moved from elsewhere, but from a standpoint of polish it might actually look a bit better to do this as adjusting the `alpha` rather than `isHidden`, which would also allow you to use `UIView.animate` blocks with a delay instead of hitting the dispatch queue directly. 

Definitely optional though.

prasanthu: Will do that later. Will make a card.

PR #227, opened 2019-05-23
Feature/esim carousel

havard024:
- insert esim carousel between esim on boarding and esim pending download


ostelco-ios-client/Assets/Assets.xcassets/instructionsESimContinue.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "screenshot1.png",
```
designatednerd: Can you please export images from Zeplin as PDF rather than 3 pngs? If you take a look at, for example `newWindowIcon`, you can see how to set up a single universal vector PDF, which saves space both at install time and in the repo.

ostelco-ios-client/ViewControllers/ESim/ESIMInstructionsViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// Has to be set up through `prepareForSegue` when this VC is loaded
+    // swiftlint:disable:next implicitly_unwrapped_optional
+    private var pageController: UIPageViewController!
+    
+    @IBOutlet private var primaryButton: PrimaryButton!
+    @IBOutlet private var lastPageLabel: BodyTextLabel!
+    
+    private lazy var dataSource: PageControllerDataSource = {
+        let pages = ESIMPage.allCases.map { $0.viewController }
+        print(pages)
```
designatednerd: do you still need this print statement?

ostelco-ios-client/ViewControllers/ESim/ESIMPageViewController.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+    
+    private func setTextLabel(_ label: OstelcoLabel, _ parts: [String]) {
+        switch parts.count {
+        case 1:
+            label.text = parts[0]
+        case 2:
+            label.setFullText(parts[0], withBoldedPortion: parts[1])
+        default:
+            break
```
designatednerd: I would really, really strongly recommend separating these into `text` and optional `boldedText` options rather than passing them as an array - you don't need to have arbitrary array of things to be bolded, just the binary Is Anything Bolded Y/N

OstelcoStyles/OstelcoLabel.swift
```diff
@@ -47,6 +47,26 @@ open class OstelcoLabel: UILabel {
         super.prepareForInterfaceBuilder()
         self.commonInit()
     }
+    
+    public func setFullText(_ fullText: String, withBoldedPortion boldedPortion: String) {
```
designatednerd: Something to be aware of when pulling this into the superclass: If  the label's `appFont` is already bold, this won't actually have any effect. 

It's fine to leave it for now though - I'll likely change this up to allow passing in attributes so we can use this for the linky text, and if it takes arbitrary attributes, then you can easily pass in some other differentiating attributes if the font is bold.

designatednerd:
Frak. LMK if you need a hand resolving the storyboard conflicts.

ostelco-ios-client/Assets/Assets.xcassets/instructionsESimContinue.imageset/Contents.json
havard024: Will do!

ostelco-ios-client/ViewControllers/ESim/ESIMInstructionsViewController.swift
havard024: of course not... I'll remove it

ostelco-ios-client/ViewControllers/ESim/ESIMPageViewController.swift
havard024: @designatednerd I agree, it felt a bit hacky when I did it, not sure quite sure how to change the logic to use a binary to check if anything is bolded or not.
My initial thought was to return an Attributed String directly, then at least I wouldn't need to return multiple values from each variable that can have a bolded part. Or I could have a separate variable containing the boldedPortion for both top and bottom text.

I could also use a tuple:
```
var bottomText: (String?, String?) { <-- could also be named tuple?
        switch self {
        case .instructions:
            return ("""
            Open the email on
            another device
            """, "another device")
       .... all the other cases ...
       }
}
```

and then set it as follows:

```
let (bottomText, boldedPortion) = esimPage.bottomText
if let bottomText = bottomText {
          if let boldedPortion = boldedPortion {
                  bottomTextLabel.setFullText(bottomText, withBoldedPortion: boldedPortion)
          } else {
                  bottomTextLabel.text = bottomText
          }
}
```

Or create an extra variable to hold optional bolded text for both top and bottom:

```
var bottomText: String? {
        switch self {
        ...
        }
}
var bottomTextBoldedPortion: String? {
        switch self {
        ...
        }
}
```

and set it as follows:

```
if let bottomText = esimPage.bottomText {
        if let boldedPortion = esimPage.bottomTextBoldedPortion {
                bottomTextLabel.setFullText(bottomText, withBoldedPortion: boldedPortion)
        } else {
                bottomTextLabel.text = bottomText
        }
}
```

OstelcoStyles/OstelcoLabel.swift
havard024: I will admit that this is me being lazy, in zeplin, the text is actually marked as body text, which is the class that contained the `setFullText`. So I will move the `setFullText` back and change the class I use for both top and bottom text label.

For future references, would it be an idea to create a subclass of `OstelcoLabel` where we could move the `setFullText` eg `AttributedOstelcoLabel`, then for each label that requires anything with attributed string, subclass from the `AttributedOstelcoLabel` instead of `OstelcoLabel`?

havard024:
> Frak. LMK if you need a hand resolving the storyboard conflicts.

Thanks, I'll give it a shot first!

ostelco-ios-client/Assets/Assets.xcassets/instructionsESimContinue.imageset/Contents.json
havard024: @designatednerd are we going to use gifs / mp4 at a later point for these images?

The two first images are available as pdf, but the two last are not according to zeplin
`This asset contains a ‘bitmap’ image inside. So no vector assets is generated.`

Just wondering, before I bug our designer.

ostelco-ios-client/ViewControllers/ESim/ESIMPageViewController.swift
havard024: After looking at #231 I think I get what you where talking about, if I change the return value to a struct which contains text and optional bolded text then I could add a function in the struct to check if it has a bolded part

designatednerd: That could work although it's probably easier in the end to just `if let` the bolded portion when you're looking at it since you're going to need to unwrap that anyway

OstelcoStyles/OstelcoLabel.swift
designatednerd: The changes in #231 make this more general and allow it to be used with any label, so if you want to just leave this I'll deal with it in merge funtimes

ostelco-ios-client/Assets/Assets.xcassets/instructionsESimContinue.imageset/Contents.json
designatednerd: Re: gifs - The designs all appear to be static images. 

Re: vectors - yeah I saw that on a couple things yesterday as well. I would say leave notes in the Zeplin document and see if there's a way to get that output as a vector image. You don't need to wait for him to finish that to merge this though.

PR #228, opened 2019-05-23
Feature/upgrade jumiosdk 3.1.2

prasanthu:
Started seeing an issue which was fixed in latest version

----- Mail from their support-----
In general the Error F01207 denotes low memory issue, means in situations of low memory available on the device the scanning process is required to stop and this error is displayed. 

We are aware of the occurrence of this specific error and have deployed a fix with the updates. The fix was deployed with version 3.1.0 However, we always advise to use the most updated version:
https://github.com/Jumio/mobile-sdk-ios


ostelco-ios-client/ViewControllers/EKYC/NRCIVerifyViewController.swift
```diff
@@ -81,7 +81,7 @@ extension NRCIVerifyViewController: NetverifyViewControllerDelegate {
     
     func createNetverifyController() {
         // Prevent SDK to be initialized on Jailbroken devices
-        if JMDeviceInfo.isJailbrokenDevice() {
+        if JumioDeviceInfo.isJailbrokenDevice() {
             print("Will not allow this from a jailbroken device")
```
designatednerd: should we have some kind of error screen for this @oisinzimmermann?

prasanthu:
> Is this version of the SDK still so gigantic that it can't be checked into GitHub? Might be nice to see if they fixed that so we don't have to keep pulling the whole cocoapods repo every single time we build

Same Error:

designatednerd:
Where's the :crying-blood: emoji when you need it?

Ugh, thanks for checking, at least.

ostelco-ios-client/ViewControllers/EKYC/NRCIVerifyViewController.swift
prasanthu: That should be a different PR

PR #229, opened 2019-05-23
Gif Videos + some asset cleanup

designatednerd:
In this PR: 
- Added gif videos and the ability to set them up and play them back
- Tested that the enum I'm using to access gif videos has actual files and that all files have enum values
- Updated several existing spots to use gif videos instead of static images
- Deleted unused and/or swapped out all assets which could be moved to vectors


OstelcoStyles/LoopingVideoView.swift
```diff
@@ -0,0 +1,98 @@
...
+
+import AVFoundation
+import UIKit
+
+/// A view designed to play looping videos to look like GIFs
+/// without the horrendous memory overhead of actual GIFs
+@IBDesignable
+public class LoopingVideoView: UIView {
+    
+    private lazy var gifPlayer = AVQueuePlayer()
```
prasanthu: Are we play GIF or MP4s ? If its MP4 change the name

OstelcoStyles/LoopingVideoView.swift
```diff
@@ -0,0 +1,98 @@
...
+        
+        return playerLayer
+    }()
+    private(set) var playerLooper: AVPlayerLooper?
+    
+    // Note: This label only shows in Interface Builder so you can tell where the view will go.
+    private lazy var ibPlaceholder: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.text = "Gif goes here!"
```
prasanthu: Should this be an icon instead of label ? Something like this: https://www.flaticon.com/premium-icon/film_1146203

OstelcoStyles/LoopingVideoView.swift
designatednerd: we are playing mp4s of gifs 🤷‍♀

I suppose `loopingVideoPlayer` would probably be more accurate though

OstelcoStyles/LoopingVideoView.swift
designatednerd: Normally I'd be with you on this, but since this is only there for Interface Builder as a placeholder (and would never be shown to an actual user since it's *only* ever accessed in `prepareForInterfaceBuilder`, it'd add weight to the `.ipa` without actual benefit to the user. I think the text is sufficient as a placeholder for building things out in IB.

designatednerd:
Merging per prior approval

PR #230, opened 2019-05-23
Use UIView animate to show the top messages.

prasanthu:


ostelco-ios-client/ViewControllers/Home/HomeViewController.swift
```diff
@@ -48,12 +48,21 @@ class HomeViewController: ApplePayViewController {
...
+            options: .curveEaseIn,
+            animations: {  [weak self] in
+                self?.hideMessages()
+            },
+            completion: nil)
+    }
+
+    func hideMessages() {
+        welcomeLabel.alpha = 0.0
+        messageLabel.alpha = 0.0
```
designatednerd: FWIW I'd just shove these into the `animations` block but if you'd rather leave 'em here, I'm good with that. Probably worth making this method `private` if you do though

ostelco-ios-client/ViewControllers/Home/HomeViewController.swift
```diff
@@ -149,11 +153,16 @@ class HomeViewController: ApplePayViewController {
...
+    }
+
+    @objc func didPullToRefresh() {
+        hideMessages()
+        refreshBalance()
     }
 
     @IBAction private func buyDataTapped(_ sender: Any) {
-       if hasSubscription {
+        if hasSubscription {
```
designatednerd: Is this whitespace change intentional to fix misalignment, or did this sneak in?

prasanthu: intentional

PR #241, opened 2019-06-05
Coordinator ALL The onboarding!

designatednerd:
This is a combination of a LOT of stuff. Apologies for the length of the PR.

- Move everything in Onboarding to using coordinators
- Make each coordinator have a `Destination` enum to allow routing to be calculated 
  - Added equatable implementations in tests for enums that have a case with an associated value, since those can't be automatically generated
- Got rid of the embedded table view controller for country in favor of just having a table view
- Added a mock location controller for testing
- Manually tested: 
  - Root coordinator up to showing esim
  - Email coordinator
  - Sign up coordinator
  - Country coordinator
  - Singapore EKYC coordinator
     - SingPass flow up to logging into Singpass
- Automated tested:
  - Root Coordinator
  - Email coordinator
  - Sign up coordinator
  - Country coordinator
  - EsimCoordinator

To do - strongly recommend doing before this is merged:
- Manually test: 
  - [ ] Esim Coordinator
  - [ ] Default EKYC Coordinator (have to disable hard-coding to singapore here)
  - [ ] Singapore EKYC Coordinator
     - [ ] NRIC / Jumio / Address flow
     - [ ] SingPass flow after actually getting address from singPass (when I was testing this was refusing to work)

To Do (can wait until after merge): 
- Fix issue where logging out or starting again from the onboarding flow won't re-show the sign up stuff - this is related to how the onboarding nav is being shown. Workaround is to force quit the app (or just rebuild if you're debugging)
- Automated test
  - Default EKYC coordinator
  - Singapore EKYC coordinator


ostelco-core/Models/SimProfile.swift
```diff
@@ -14,9 +14,19 @@ public enum SimProfileStatus: String, Codable {
     case NOT_READY
 }
 
-public struct SimProfile: Codable {
+public struct SimProfile: Codable, Equatable {
```
designatednerd: For this one since `SimProfileStatus` and `String` are both `Equatable`, the compiler auto-generates the implementation.

ostelco-ios-client/Coordinators/EKYC/DefaultEKYCCoordinator.swift
```diff
@@ -0,0 +1,113 @@
...
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import ostelco_core
+import PromiseKit
+import UIKit
+
+class DefaultEKYCCoordinator: EKYCCoordinator {
+    
+    static func coordinator(for country: Country,
```
designatednerd: I tried to put this on the `EKYCCoordinator` protocol, but the compiler warned his has to be a static func on a concrete type or it can't actually generate a coordinator. Figured default was probably the better of the two to use as a starting point.

ostelco-ios-client/Coordinators/EKYC/SingaporeEKYCCoordinator.swift
```diff
@@ -0,0 +1,222 @@
...
+        case .PENDING:
+            guard
+                let jumio = region.kycStatusMap.JUMIO,
+                let addressAndPhoneNumber = region.kycStatusMap.ADDRESS_AND_PHONE_NUMBER,
+                let nricFin = region.kycStatusMap.NRIC_FIN,
+                let myInfo = region.kycStatusMap.MY_INFO else {
+                    return .goBackAndChooseCountry
+            }
+            
+            switch (jumio, addressAndPhoneNumber, nricFin, myInfo) {
```
designatednerd: would appreciate as many extra sets of eyes on this bit of logic as possible @havard024 @prasanthu

ostelco-ios-client/Coordinators/RootCoordinator.swift
```diff
@@ -35,21 +53,81 @@ class RootCoordinator {
...
+                return self.handleRegionResponse(region)
+            }
+            // Recover allows us to check for an error but continue the chain
+            .recover { error -> Promise<RootCoordinator.Destination> in
+                switch error {
+                case APIHelper.Error.jsonError(let requestError):
+                    if requestError.httpStatusCode == 404 {
+                        return .value(.signUp)
+                    } // else, keep going.
+                default:
```
designatednerd: There was another case here before but we don't need it because the API we're using here returns the `jsonError` type

ostelco-ios-client/Protocols/LocationCheckingViewController.swift
```diff
@@ -1,84 +0,0 @@
-//
```
designatednerd: Logic here got moved to the Country Coordinator

ostelco-ios-client/ViewControllers/EKYC/NRCIVerifyViewController.swift
```diff
@@ -1,173 +0,0 @@
-//
-//  NRCIVerifyViewController.swift
```
designatednerd: This just got its name changed to the correct acronym (NRIC) - apparently there were enough changes it's showing up as a delete and add rather than a diff

dev-ostelco-ios-clientTests/CoordinatorTests/RootCoordinatorTests.swift
```diff
@@ -0,0 +1,167 @@
...
+class RootCoordinatorTests: XCTestCase {
+    
+    lazy var testUserManager = MockUserManager()
+    
+    lazy var testCoordinator = RootCoordinator(window: UIWindow(),
+                                               userManager: self.testUserManager)
+    
+    override func setUp() {
+        super.setUp()
+        UserDefaultsWrapper.clearAll()
```
sgoodwin: What if we made it so we could feed a fresh fake user defaults in for each test? This way seems easy to mess up and get inconsistent test results.

dev-ostelco-ios-clientTests/CoordinatorTests/RootCoordinatorTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        self.tearDownStubbing()
+        APIManager.shared.primeAPI = PrimeAPI(baseURL: APIManager.shared.baseURLString, tokenProvider: UserManager.shared)
+        UserDefaultsWrapper.clearAll()
+        super.tearDown()
+    }
+    
+    func testNoLoggedInUserKicksToLogin() {
+        self.testUserManager.mockHasCurrentUser = false
+        
+        guard let destination = self.testCoordinator.determineDestination().awaitResult(in: self) else {
```
sgoodwin: awaitResult works without needing to use a test expectation to wait? handy

ostelco-ios-client/AppDelegate.swift
```diff
@@ -122,10 +122,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if EmailLinkManager.isSignInLink(incomingURL) {
             EmailLinkManager.signInWithLink(incomingURL)
                 .then {
-                    UserManager.shared.getDestinationFromContext()
+                    self.rootCoordinator.determineDestination()
                 }
                 .done { [weak self] destination in
-                    self?.rootCoordinator.navigate(to: destination, from: nil)
+                    self?.rootCoordinator.navigate(to: destination, from: nil, animated: true)
```
sgoodwin: So the coordinator figured out where should be next which is handy to keep up with, but here we're relying on the AppDelegate to actually go there next. In theory the root coordinator already knows we need to go somewhere and could direct that without the app delegate asking it to?

ostelco-ios-client/Coordinators/CountryCoordinator.swift
```diff
@@ -0,0 +1,151 @@
...
+
+import ostelco_core
+import PromiseKit
+import UIKit
+
+protocol CountryCoordinatorDelegate: class {
+    func countrySelectionCompleted(with country: Country)
+}
+
+class CountryCoordinator {
```
sgoodwin: Pretty understandable stuff. <3

ostelco-ios-client/Coordinators/EKYC/SingaporeEKYCCoordinator.swift
sgoodwin: Is there a diagram of this somewhere? Seems like important business logic stuff.

ostelco-ios-client/ModelControllers/UserManager.swift
```diff
@@ -63,75 +63,6 @@ class UserManager: TokenProvider {
         OnBoardingManager.sharedInstance.region = nil
     }
     
-    func getDestinationFromContext() -> Promise<PostLoginDestination> {
```
sgoodwin: Let it burn! 🔥

dev-ostelco-ios-clientTests/CoordinatorTests/RootCoordinatorTests.swift
designatednerd: It uses expectations [under the hood](https://github.com/ostelco/ostelco-ios-client/blob/develop/dev-ostelco-ios-clientTests/PromiseKit%2BTesting.swift#L14)

ostelco-ios-client/AppDelegate.swift
designatednerd: Yeah that makes sense.

ostelco-ios-client/ModelControllers/UserManager.swift
designatednerd: It's not burnt, it's just broken into pieces 😄

dev-ostelco-ios-clientTests/CoordinatorTests/RootCoordinatorTests.swift
designatednerd: The wrapper would need some architectural changes to support this - I'd say maybe make a trello card to do it later.

ostelco-ios-client/Coordinators/EKYC/SingaporeEKYCCoordinator.swift
designatednerd: Not that I've seen but maybe @havard024 @prasanthu know?

dev-ostelco-ios-clientTests/CoordinatorTests/RootCoordinatorTests.swift
sgoodwin: Oh nice.

dev-ostelco-ios-clientTests/CoordinatorTests/RootCoordinatorTests.swift
sgoodwin: I'll make a note, agreed.

ostelco-ios-client/ModelControllers/UserManager.swift
sgoodwin: Close enough!

ostelco-ios-client/Coordinators/EKYC/SingaporeEKYCCoordinator.swift
```diff
@@ -0,0 +1,222 @@
...
+            case (.PENDING, .PENDING, .PENDING, .PENDING):
+                // User hasn't selected anything yet
+                return .selectVerificationMethod
+            case (.REJECTED, _, _, _):
+                // If jumio is rejected, *everything* is rejected.
```
prasanthu: @havard024 @vihangpatil Can you confirm this ? We can have a case where Jumio is Rejected, but MyInfo and address are approved.

ostelco-ios-client/Coordinators/EKYC/SingaporeEKYCCoordinator.swift
```diff
@@ -0,0 +1,222 @@
...
+            case (_, .PENDING, _, .APPROVED):
+                // My info also needs an address submitted, but we've probably lost
+                // what we need to access their singpass address. Make the user enter it.
+                return .enterAddress
+            case (.PENDING, .PENDING, .APPROVED, _):
+                // The user has an approved NRIC. Now we need Jumio verification and an address. Start with jumio.
+                return .jumio
+            case (.PENDING, .APPROVED, .APPROVED, _):
+                // The user has an approved NRIC and address, now we need Jumio verification.
+                return .jumio
```
prasanthu: With this code path, Will go through address after Jumio (Which is already approved)?

ostelco-ios-client/Coordinators/EKYC/SingaporeEKYCCoordinator.swift
havard024: I'm going through the app manually to check if all cases are covered, as soon as I get through the flow, I'll comment if there is anything missing

designatednerd: I'm also working on some tests around this - trying to figure out how to cover every case is a bit of a nightmare

PR #242, opened 2019-06-07
Singapore EKYC is Hard. 

designatednerd:
In this PR: 
- Updated EKYC to match what I *think* are the updated cases for singapore
- Simplified things as much as I could by breaking into flows
- Added a markdown table where I'm starting to break down all the different cases. Spoiler: There are A LOT.


sgoodwin:
💚  I dunno about the business logic so much yet, but the rest seems fine.

designatednerd:
Taking 💚as an approval and merging to the main feature branch

PR #243, opened 2019-06-08
add deep link to allow login with username and password

havard024:
Allows username and password login using deep link.

You can only use this deep link to share credentials to existing users.

To share credentials of existing user:
1. Set a password for an existing user by resetting the password for an existing user in firebase console
2. Construct the deep link https://dl-dev.oya.world/login-with-username-and-password?username=[email]&password=[password]
3. Make sure the app is already closed to prevent routing logic issues
4. enter the url from 2 to a browser on the phone
5. then manually open the app from the status bar in the browser if the app doesn't open automatically


sgoodwin:
In theory it shouldn't need to be a step to close the app first? Or is this just because the big routing PR is being worked on?

havard024:
> In theory it shouldn't need to be a step to close the app first? Or is this just because the big routing PR is being worked on?

Usually no, though I did experience once that the deep link did not work, or rather, it did not present the correct VC. I don't remember exactly where, but I think it was because of this line:
https://github.com/ostelco/ostelco-ios-client/blob/970d4dc9e6c6eab8764b28bfcf5f8a39e4c85d5a/ostelco-ios-client/Coordinators/RootCoordinator.swift#L206

Though when I think about it, that's actually fine, if we want to show the same screen as you already were on. I guess my thinking there was that, since there was no indication that you logged in, something was wrong, I guess if we could add some visual feedback that you successfully logged in with another user, at least you would be notified about it. (If not it might seem like the deeplink is not working)

PR #244, opened 2019-06-08
fix shake to open help menu

havard024:
The existing code did trigger when you shook the phone.

### DIFF
--- a/ostelco-ios-client/Extensions/UIWindow+ShakeToOpenHelp.swift
+++ b/ostelco-ios-client/Extensions/UIWindow+ShakeToOpenHelp.swift
@@ -9,8 +9,8 @@
 extension UIWindow {
     open override func motionEnded(_ motion: UIEvent.EventSubtype, with event: UIEvent?) {
         if motion == .motionShake {
-            guard let vc = UIApplication.shared.typedDelegate.rootCoordinator.topViewController else {
-                //
+            guard let vc = self.rootViewController else {
+                // There's no view controller to show.
                 debugPrint("Can't show need help action sheet, there's no view controller to show.")
                 return
             }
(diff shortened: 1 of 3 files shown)


ostelco-ios-client/Extensions/UIWindow+ShakeToOpenHelp.swift
```diff
@@ -0,0 +1,27 @@
...
+//  ostelco-ios-client
+//
+//  Created by mac on 6/8/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+extension UIWindow {
+    open override func motionEnded(_ motion: UIEvent.EventSubtype, with event: UIEvent?) {
+        if motion == .motionShake {
+            guard let vc = UIApplication.shared.typedDelegate.rootCoordinator.topViewController else {
```
sgoodwin: I think you could do this with the window's `rootViewController` instead of needing to dig so deep into the properties of other objects.

havard024: I tried with
`guard let vc = self.window?.rootViewController else {
                // There's no view controller to show.
                return
            }
`
without any luck

sgoodwin: In this case self *is* the window, so just `self.rootViewController` should be sufficient. Unless your issue was something other than a compiler problem?

havard024: thanks! That worked

ostelco-ios-client/Extensions/UIWindow+ShakeToOpenHelp.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  UIWindow+ShakeToOpenHelp.swift
+//  ostelco-ios-client
+//
+//  Created by mac on 6/8/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+extension UIWindow {
+    open override func motionEnded(_ motion: UIEvent.EventSubtype, with event: UIEvent?) {
```
designatednerd: @havard024 @sgoodwin I would really, really strongly recommend against overriding a function in an extension - you can wind up with undefined or unpredictable behavior. With the app delegate method, did it work if you used the RootCoordinator's top view controller instead of just the root view controller?

PR #245, opened 2019-06-10
Myinfo fix for old prod client

havard024:


ostelco-core/Models/MyInfoDetails.swift
```diff
@@ -92,7 +92,7 @@ public struct MyInfoDetails: Codable {
         case _name = "name"
         case _dob = "dob"
         case _email = "email"
-        case address = "regadd"
+        case address = "mailadd"
```
designatednerd: I think we should probably be passing a flag so that we can have this be `#if`-`#else`'d based on whether we're on dev or prod build. Otherwise this'll work on one but not the other.

ostelco-core/Network/Request.swift
```diff
@@ -47,6 +47,7 @@ public struct Request {
         
         var request = URLRequest(url: url)
         request.httpMethod = self.method.rawValue
+        request.timeoutInterval = 180.0
```
designatednerd: do we want every request to have a 5 minute timeout, or just some requests?

havard024: This was just an attempt to get something ready for regulator meeting. We shouldn't keep it.
There's only one endpoint that can take > 60 seconds (observed both in prod and dev) and that is this endpoint `GET /regions/sg/kyc/myInfo/{authorisationCode}`

I really want to figure out why it can take that long, before we `solve` it by increasing the request timeout

ostelco-core/Models/MyInfoDetails.swift
havard024: I think this renaming needs to happen in dev env also

ostelco-core/Network/Request.swift
designatednerd: cool - maybe revert this for the time being then

havard024:
Closing this since the fix we want has been added to develop already

PR #247, opened 2019-06-11
Show help menu when user shakes device.

sgoodwin:
This was not triggering in the App Delegate, but the Splash controller
is always the app's root view controller. Another solution was to add
this method as an extension to UIWindow, but override methods in an
extension are dangerous (and create warnings).


designatednerd:
@sgoodwin Getting you added to the org - in the future you'll likely need to do PRs straight from the main repo since we can't build with secrets on PRs from other Repos for security reasons.

PR #248, opened 2019-06-11
CodeGen Images and GifVideo enums

designatednerd:
Changes aren't quite as bad as they seem here - the `SPM.xcodeproj` file takes up most of them. 

In this PR: 
- Update swift PM dependency to 0.3.0 (we can't update it higher until we support Swift 5.1, unfortunately, so those two stupid warnings will be with us until then)
- Use Pre-Build script to code-gen enums for images in the asset catalog and videos in the `gifMP4s` folder.


designatednerd:
@havard024 @prasanthu @sgoodwin This is now ready for review

PR #251, opened 2019-06-12
Enable US during debugging, update parameter name for MyInfo

designatednerd:
This slightly duplicates #245, but this should get us to where we need to be for SingPass and US EKYC debugging.


sgoodwin:
Looks fine 🍏

PR #254, opened 2019-06-13
Feature/feature flags

havard024:
This PR shows how we could implement feature flags using firebase remote config.

There are two example videos, one showing how remote config works by printing a welcome message on startup and when app enters foreground.

Second example shows how we could use remote config to set which countries require location check.

Example video:
https://www.loom.com/share/18943bfebf7743848daf2c1334810701

Example video on how to use remote config to set which countries require location check:
https://www.loom.com/share/f1fbe4c8dc694e469f05b719a2a31cf0

Remote Config is a firebase product where you can set up to 2000 individual parameters.
Each parameter is defined as a string
They have a inbuilt json editor if you want more complex data structures, it's still stored as a string and needs to be parsed from string to array / dict in the client.
You can also set conditions to parameters, meaning only clients that fulfills the condition will receive the parameters. (Conditions is based on the same filters you use for Firebase Analytics, user audience, country etc etc).
It is also possible to set default values bundled with the client application.

Read more about Remote Config here: https://firebase.google.com/docs/remote-config


havard024:
Getting kinda old.

PR #264, opened 2019-06-25
Make shake to help as new root for the whole app.

havard024:
Refactored shake to help to its own view controller and made the new view controller the new root for the whole app, because the shake to help was handled by the splash view controller which should not be responsible for that as well.


sgoodwin:
Is this waiting on something?

PR #266, opened 2019-07-01
add initial test setup for routing logic

havard024:
initial test setup for routing logic, the test setup is incomplete as of now, but this would make it easier for us to get an overview of the complex routing as well as make it easier for future changes


havard024:
Closing this since same code with working tests can be found in #267

PR #277, opened 2019-07-10
Feature/myinfo v3

prasanthu:
Myinfo V3 has changed the structure of returned JSON. Sample return values can be found in the documentation here: https://www.ndi-api.gov.sg/assets/lib/trusted-data/myinfo/specs/myinfo-kyc-v3.0.2.html#operation/getperson


ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -303,6 +303,13 @@ extension OnboardingCoordinator: MyInfoSummaryDelegate {
     func verifiedSingPassAddress() {
         advance()
     }
+
+    func failedToLoadMyInfo() {
+        if localContext.selectedVerificationOption == .singpass {
```
sgoodwin: I don't think there's a need for the conditional here.

prasanthu: done

ostelco-core/Network/Headers.swift
```diff
@@ -13,6 +13,7 @@ public enum HeaderKey: String {
     case authorization = "Authorization"
     case contentType = "Content-Type"
     case testing = "Testing"
+    case mode = "x-mode"
```
havard024: I added a different way of setting the x-mode heade, using env variables passed when building a target. I'm using it for GraphQL here: https://github.com/ostelco/ostelco-ios-client/blob/e9c9a0990549c08990046dcbd5e889adf3409ed4/ostelco-core/Network/PrimeAPI.swift#L631

I don't have a strong opinion on how we do it, though we might want to do it the same way, or maybe it's not a big deal since we are about to replace the legacy APIs with GraphQL

prasanthu: We can do this later in a different PR

PR #280, opened 2019-07-11
Lock SPM minor version.

havard024:
Lock SPM minor version since I experienced issues building the project with SPM version 0.4.0.

scripts/SPM/swift build
'SPM' /Users/mac/github/xcode/ostelco-ios-client/scripts/SPM: error: product dependency 'Utility' not found
warning: dependency 'SwiftPM' is not used by any target


sgoodwin:
Or we can fix the issue for newer SPM. Let's not just merge all these things without some discussion eh?

PR #281, opened 2019-07-15
Feature/migrate to graphql

havard024:
- Updated Apollo from 0.10.1 to 0.11.1 to be able to set Authorization header using HTTPNetworkTransportPreflightDelegate using a custom networkTransport for ApolloClient
- Added extension helpers for existing models to be able to convert the GraphQL models to our existing models, this is only during first step of migration over to GraphQL so that we don't do too much at once.
- The REST API and GraphQL schema does not always map when it comes to Optional / Required variables. You can see this in the extensions added for our models where we have to force unwrap. Ideally we update the GraphQL schema to reflect the Required variables we expect in the REST API.
- Using GraphQL fragments to promote code reuse see context.graphql and products.graphql for examples together with the corresponding queries in PrimeAPI.
- Added env parameter in xcode for the dev target to set the `X-Mode` header, the latest schema is currently deployed in canary, thus you would need to set the env parameter for xMode=canary on. 

Remaining legacy GET APIs not yet in the GraphQL schema to be converted at a later point:
- `GET /regions/sg/kyc/dave/{nricFinId}`
- `GET /regions/sg/kyc/myInfo/{authorisationCode}`
- `GET /regions/sg/kyc/myInfoConfig`
- `GET /customer/stripe-ephemeral-key`
- `GET /regions/{regionCode}/simProfiles/{iccId}/resendEmail`


ostelco-core/Network/PrimeAPI.swift
```diff
@@ -444,4 +516,23 @@ open class PrimeAPI: BasicNetwork {
                 self.performRequest($0)
             }
     }
+    
+    private func getToken() -> PromiseKit.Promise<String> {
```
havard024: Helper function so we can set Authorization header when using GraphQL endpoints

PR #292, opened 2019-07-19
Remove the onboarding manager.

sgoodwin:
It wasn't accomplishing much except giving everyone global access to a
variable they should not have. Global variables cause issues, in this
case it caused the app to provision extra eSim profiles and would cause
location checking to work incorrectly.


havard024:
pushed this branch with a new prefix to make our CI build and test #293

PR #293, opened 2019-07-22
Remove the onboarding manager.

havard024:
It wasn't accomplishing much except giving everyone global access to a
variable they should not have. Global variables cause issues, in this
case it caused the app to provision extra eSim profiles and would cause
location checking to work incorrectly.


sgoodwin:
But it's not a feature, it's fixing a bug. This prefix thing is pretty silly.

havard024:
#295

PR #308, opened 2019-08-01
Started refactoring and found a bug.

sgoodwin:
I started by dumbing down the MyInfoSummaryViewController which was
still making API calls on its own. I found that the region we update was
hard-coded once we finish with that step which is not what we want. I
also found we were incorrectly ignoring dynamic links when a pending
email was not present even though the email did not matter by the time
we are handling a dynamic link. This
corrects the issues.

### DIFF
--- a/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
+++ b/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
@@ -121,7 +121,7 @@ class OnboardingCoordinator {
             selectEKYCMethod.delegate = self
             navigationController.setViewControllers([selectEKYCMethod], animated: true)
         case .singpass:
-            singpassCoordinator = SingPassCoordinator(delegate: self)
+            singpassCoordinator = SingPassCoordinator(delegate: self, primeAPI: primeAPI)
             singpassCoordinator?.startLogin(from: navigationController)
         case .verifyMyInfo(let code):
             let verifyMyInfo = MyInfoSummaryViewController.fromStoryboard()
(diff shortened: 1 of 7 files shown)


ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -377,23 +377,49 @@ extension OnboardingCoordinator: SingPassCoordinatorDelegate {
...
+        let spinnerView = controller.showSpinner(loadingText: NSLocalizedString("Loading your data from SingPass...", comment: "Loading text after user approves SingPass"))
+        primeAPI.loadSingpassInfo(code: code)
+        .ensure {
+            controller.removeSpinner(spinnerView)
+        }
+        .done { myInfoDetails in
+            completion(myInfoDetails)
+        }
+        .catch { error in
+            ApplicationErrors.log(error)
```
prasanthu: This changes the way we handle errors. Last implementation will 
1) Show error
2) Go back to Select verification method screen.

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -377,23 +377,51 @@ extension OnboardingCoordinator: SingPassCoordinatorDelegate {
...
+            controller.showGenericError(error: error) { [weak self] (_) in
+                self?.advance()
```
prasanthu: You are not doing 
```
self?.localContext.selectedVerificationOption = nil
self?.localContext.myInfoCode = nil
```
before calling advance. It needs to go back to select identity verification screen

PR #310, opened 2019-08-09
Feature/sign in with apple

prasanthu:
Add Sign In with Apple to the Dev target.

1) Sign-in is the last step of Login Carousel
2) Email entry and waiting for dynamic link is not necessary
3) Dropped emailEntry & checkYourEmail stages

Pending:
-The unused storyboards and classes will be removed in another PR
-Private relay emails are not getting delivered (Issue in the backend)
-Add keys for prod target in server


ostelco-ios-client/ViewControllers/Login/LoginViewController.swift
```diff
@@ -53,28 +57,46 @@ class LoginViewController: UIViewController {
...
         } else {
             primaryButton.setTitle(NSLocalizedString("Next", comment: "Action button in Carousel"), for: .normal)
+            primaryButton.isHidden = false
+            buttonStackView.isHidden = true
         }
     }
-    
+
+    func setupProviderLoginView() {
+        authorizationButton = ASAuthorizationAppleIDButton(type: .signIn, style: .whiteOutline)
```
havard024: I got an compiler error here, not sure why, but locally the `signUp` type does not exist for me

Screen Shot 2019-08-16 at 1 43 39 PM

ostelco-core/ModelControllers/UserDefaultsWrapper.swift
```diff
@@ -14,6 +14,7 @@ public struct UserDefaultsWrapper {
     private enum Key: String, CaseIterable {
         case pendingEmail = "PendingEmail"
```
havard024: I guess we could remove this, if it's only used with our firebase login

ostelco-ios-client/ViewControllers/Login/LoginViewController.swift
prasanthu: I see a different definition 
Screenshot 2019-08-16 at 14 00 33

ostelco-core/ModelControllers/UserDefaultsWrapper.swift
prasanthu: My plan was to cleanup all unused classes, enums and storyboards in a different PR

havard024: That makes sense 👌

ostelco-ios-client/ViewControllers/Login/LoginViewController.swift
havard024: No issues, I didn't have the latest xcode-beta

PR #311, opened 2019-08-12
Prevent multiple eSim profile creation.

sgoodwin:
If you get through the onboarding and hit continue after seeing the eSim
instructions multiple times, you can get multiple sims. This prevents
that by making the button un-tappable after you tap it once.

### DIFF
--- a/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
+++ b/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
@@ -455,11 +455,16 @@ extension OnboardingCoordinator: ESIMOnBoardingDelegate {
 }
 
 extension OnboardingCoordinator: ESIMInstructionsDelegate {
-    func completedInstructions() {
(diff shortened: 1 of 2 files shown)


ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -455,11 +455,16 @@ extension OnboardingCoordinator: ESIMOnBoardingDelegate {
 }
 
 extension OnboardingCoordinator: ESIMInstructionsDelegate {
-    func completedInstructions() {
+    func completedInstructions(_ controller: ESIMInstructionsViewController) {
         localContext.hasSeenESIMInstructions = true
         
+        let spinner = controller.showSpinner()
```
prasanthu: By calling`controller.showSpinner()`, it will prevent user from clicking the continue twice, right ?

sgoodwin: Indeed. Showing the spinner blocks all of the UI so you can't tap on anything.

PR #316, opened 2019-08-15
Update pod JumioMobileSDK/Netverify from 3.1.2 to 3.3.0.

havard024:
I want to test the jumio flow before we merge this, forgot my passport today, so won't be able to test before tomorrow...


havard024:
Tested and verified that Jumio works as before.

PR #319, opened 2019-08-15
Add utility to convert ISO 3166 2 to 3 and vice versa.

havard024:
Solution soly based on SO: https://stackoverflow.com/questions/11576947/ios-convert-iso-alpha-2-to-alpha-3-country-code.


havard024:
There have been some thoughts about how we could do this in a simpler way, possibly better way, since we don't need to support all countries as of now. And the list of countries will grow very slowly. There will be a new PR with a new suggestion on how to solve this.

PR #321, opened 2019-08-19
Remove unused storyboards, classes and enums

prasanthu:
Sign-In-With-Apple has made few of the classes obsolete. This patch remove those


ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -188,7 +188,8 @@ class OnboardingCoordinator {
 }
 
 extension OnboardingCoordinator: LoginDelegate {
-    func signedIn(authCode: String, contactEmail: String?) {
+    func signedIn(controller: UIViewController, authCode: String, contactEmail: String?) {
+        controller.showSpinner()
```
sgoodwin: You'll need to remove the spinner in the case where sign in fails and we can't go to the next screen.

prasanthu: Added `controller.removeSpinner(spinnerView)` in `catch`

PR #323, opened 2019-08-20
Change error for Authorization failures.

prasanthu:


sgoodwin:
Is there only the one reason this will fail?

prasanthu:
> Is there only the one reason this will fail?

Network Connection failures should be captured elsewhere.

prasanthu:
> Is there only the one reason this will fail?

@sgoodwin You are right, I missed the error handling in `ASAuthorizationControllerDelegate`

PR #330, opened 2019-08-23
Global Location Check: Showing country on Home view.

sgoodwin:
This adds support for showing the current country once you're done with onboarding.

This also fixes an issue where it would think you were done with on boarding because it didn't have the myinfo code anymore.

### DIFF
--- a/dev-ostelco-ios-clientTests/CoordinatorTests/EdgeCasesInStageDeciderTests.swift
+++ b/dev-ostelco-ios-clientTests/CoordinatorTests/EdgeCasesInStageDeciderTests.swift
@@ -142,7 +142,7 @@ class EdgeCasesInStageDeciderTests: XCTestCase {
     
     func testUserHasCompletedJumioButGotRejected() {
         let decider = StageDecider()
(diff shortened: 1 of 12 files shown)


prasanthu:
I generated the eKYC rejection by hiding half of my driver's license with sticky notes.  There is some documentation from JUMIO (I haven't tried those) https://github.com/Jumio/implementation-guides/blob/master/netverify/netverify-test-ids.md

havard024:
I detected to issues when testing:

Given a user who's status in Norway is rejected.
After the user completes ekyc again, the app presents the oh no ekyc rejected screen immediately (because of the context = rejected), but it should show the pending verification screen.

Given a new user who signs up for norway who closes the app after completing the ekyc where the app is presenting the pending verification screen. Then while the app is closed, the ekyc is rejected, then the user opens the app again, the user is token to the ekyc process, but should actually see the ekyc failed screen first, before starting the ekyc process again.

sgoodwin:
As discussed, merging this for now. Can look at Havard's issue separately.

PR #331, opened 2019-08-28
Update homescreen after User has become a member.

prasanthu:


sgoodwin:
I think this would benefit from a protocol rather than having the BecomeMember view controller directly manipulate the home view controller. Just a thought

sgoodwin:
👍

PR #333, opened 2019-08-29
Feature/fix send qr code again

prasanthu:
We are no removing the spinner after the API request is finished.

### DIFF
--- a/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
+++ b/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
@@ -476,8 +476,6 @@ extension OnboardingCoordinator: ESIMInstructionsDelegate {
         primeAPI.loadContext()
         .then { (context) -> PromiseKit.Promise<PrimeGQL.SimProfileFields> in
             assert(context.regions.count == 1)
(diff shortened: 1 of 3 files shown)


ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -476,8 +476,6 @@ extension OnboardingCoordinator: ESIMInstructionsDelegate {
         primeAPI.loadContext()
         .then { (context) -> PromiseKit.Promise<PrimeGQL.SimProfileFields> in
             assert(context.regions.count == 1)
-            // swiftlint:disable:next empty_count
-            assert(context.regions.first!.fragments.regionDetailsFragment.simProfiles?.count == 0)
             let simProfile = RegionResponse.getRegionFromRegionResponseArray(context.regions.map({ $0.fragments.regionDetailsFragment }))?.getSimProfile()
```
prasanthu: @havard024 See my comment in the patch for this.

havard024: I agree that the assert does not make sense according how we want this to work.

PR #334, opened 2019-08-30
Use context region if no selected region exists

havard024:


sgoodwin:
What is this attempting to fix?

havard024:
Fixes this issue https://trello.com/c/1jvR0kcC/261-if-user-has-region-with-status-pending-in-context-and-starts-the-app-from-scratch-user-is-taken-to-home-screen-instead-of-jumio

PR #338, opened 2019-09-04
Make sure we have camera access before jumio starts.

havard024:
Missing design on the dont have camera access screen.


sgoodwin:
This doesn't need to be a special exception, if you make `cameraAccess` a new step and add it in before Jumio, you can then have it remove the step if it sees that the local context knows we have camera access. This'll also allow us to use it anywhere, like if we have some other ekyc system in the future that also needs the camera. Make sense?

havard024:
sure

PR #339, opened 2019-09-06
Feature/setup apple pay from buy products

prasanthu:
We are getting rid of membership from the product. So we need a way to make sure user has Apple pay. This patch will ask to setup Apple Pay (if necessary) at the time of buying a product.


ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+
+class SetupApplePayViewController: ApplePayViewController {
+
+    @IBOutlet private var buttonContainer: UIView!
+    @IBOutlet private var explanatoryCopyLabel: BodyTextLabel!
+
+    var paymentButton: PKPaymentButton?
+    var membership: Product?
+
+    lazy var linkableCopy: LinkableText = {
```
sgoodwin: Just for the record: there's no need to store this stuff anymore. The delegate method now informs which link was tapped, so we don't need to save it to match against.

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+
+        setupPaymentButton()
+    }
+
+    func setupPaymentButton() {
+        // Log information about the current Apple Pay setup.
+        let applePayError: ApplePayError? = canMakePayments()
+        switch applePayError {
+        case .unsupportedDevice?:
+            debugPrint("Apple Pay is not supported on this device")
```
sgoodwin: Do we need to do something during these errors?

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+        paymentButton.centerXAnchor.constraint(equalTo: buttonContainer.centerXAnchor).isActive = true
+        paymentButton.centerYAnchor.constraint(equalTo: buttonContainer.centerYAnchor).isActive = true
+        paymentButton.cornerRadius = 8.0
+        self.paymentButton = paymentButton
+    }
+
+    @objc func setUpButtonTapped() {
+        PKPassLibrary().openPaymentSetup()
+        // Go back to home screen.
+        dismiss(animated: true)
```
sgoodwin: Controllers generally should not be dismissing themselves. This assumes how the controller will be shown and can result in odd bugs later (especially when more than one controller is presented).

prasanthu: Are you proposing the parent to dismiss this (in this case HomeViewController )? Also should this be done via a delegate telling the parent about this event ?

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
prasanthu: No, I will remove this. The object creating this should check for these things...

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
prasanthu: Can you suggest what I do here. I was just copying the structure in the "BecomeAMemberViewController"

sgoodwin: I mean you can create and assign the text directly. There's no need to make a property for this text.

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
sgoodwin: Apple documentation recommends the controller that does the presenting handle the dismissing.

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
prasanthu: Remove `linkableCopy` property

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
prasanthu: It now uses protocols to convey the info back to the presenter

ostelco-ios-client/ViewControllers/Home/SetupApplePayViewController.swift
prasanthu: Removed

PR #342, opened 2019-09-13
Feature/remove membership

prasanthu:
We are getting rid of the "Membership" concept in the product. This PR removes all traces of membership from the App. The screen for "Become a member"  and the logic on how to present product list have been changed.

### DIFF
--- a/ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
+++ b/ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
@@ -12,7 +12,7 @@ import PromiseKit
 import Stripe
 import ostelco_core
 
-// This is the base class for both HomeViewController and BecomeAMemberViewController
+// This is the base class for both HomeViewController and SetupApplePayViewController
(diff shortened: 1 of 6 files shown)


ostelco-ios-client/ViewControllers/Home/ApplePayViewController.swift
```diff
@@ -12,7 +12,7 @@ import PromiseKit
 import Stripe
 import ostelco_core
 
-// This is the base class for both HomeViewController and BecomeAMemberViewController
+// This is the base class for both HomeViewController and SetupApplePayViewController
 // 1) Adds properties defined by the ApplePayDelegate protocol
 // 2) Adds @objc methods for PKPaymentAuthorizationViewControllerDelegate
 // This class coudn't be avoided due to the issue described in the following link.
```
la3lma: Typo :-) (couldn't).  I _strongly_ approve of starting this file with a high level comment for the class.

prasanthu: Pushed a new PR to fix this

PR #352, opened 2019-09-26
Feature/apple esim

havard024:
We have to remember to add the esim development entitlement in the development profile for our prod app when we eventually merge this.


ostelco-core/Models/StageDecider.swift
```diff
@@ -19,6 +19,7 @@ public class LocalContext {
     public var hasSeenVerifyIdentifyOnboarding: Bool
     public var selectedVerificationOption: IdentityVerificationOption?
     public var hasSeenLocationPermissions: Bool
+    public var simProfile: SimProfile?
```
sgoodwin: Why are we adding the sim profile to the local context when the Context already has it?

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -218,6 +215,52 @@ class OnboardingCoordinator {
         }
         return false
     }
+    
+    /**
+     Tries to download and install an eSIM on the device using apples eSIM APIs.
+     */
+    func addPlanForSimProfile(_ simProfile: SimProfile) -> PromiseKit.Promise<Void> {
+        let planObj = CTCellularPlanProvisioning()
```
sgoodwin: We should probably make a thing to handle this rather than add the functionality directly to the coordinator, no? The act of adding an esim I mean.

ostelco-core/Models/MyInfoDetails.swift
```diff
@@ -137,7 +137,7 @@ public struct MyInfoDetails: Codable {
             _regadd = newValue
         }
     }
-    
+        
     public let mobileNumber: MyInfoMobileNumber?
```
sgoodwin: This needs to go away. We don't collect mobile number anymore, so it'll be confusing to see this.

ostelco-core/Models/StageDecider.swift
havard024: In a multi device world, we don't really know which sim profile from the server context we should use. There could be a profile ready for download that was initiated from a different device etc etc.
To be on the safe side, we always provision a new sim profile if you enter the esim instructions flow, at least until we figure out which cases it's safe to use an existing sim profile from server context.

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
havard024: Where do you suggest we put this logic?

sgoodwin: I mean contain it into it's own struct/class where we can give it a sim and ask it to do the work. The onboarding coordinator's job is to handle changing steps in the on boarding flow, not to install esims.

havard024: I've moved the addPlan to a separate file.

ostelco-ios-client/ModelControllers/ESimManager.swift
```diff
@@ -0,0 +1,42 @@
...
+//  Created by mac on 9/27/19.
+//  Copyright © 2019 mac. All rights reserved.
+//
+
+import CoreTelephony
+import PromiseKit
+import ostelco_core
+
+class ESimManager {
+    static let shared = ESimManager()
```
sgoodwin: Yes, like this! Why is it a singleton though?

havard024: That's they only pattern I know 😎

sgoodwin: Hmm, maybe we can leave it for now and I'll change it with the stage decider split.

PR #355, opened 2019-09-30
Make the first shift for multi-region.

sgoodwin:
Now the onboarding is split into two segments. First the user goes
through picking a nickname and seeing the legal stuff, then the user
will see the home view. They'll do the part for EKYC and eSIM
downloading after they select a country.

This covers the first phase, removing the EKYC and eSIM portions from
the intial onboarding. Once the new country-selection is built, I'll
iron out the last bits for EKYC and eSIM portions.


dev-ostelco-ios-clientTests/CoordinatorTests/SingaporeUserHappyFlowWithSingPassStageDeciderTests.swift
```diff
@@ -66,128 +66,85 @@ class SingaporeUserHappyFlowWithSingPassStageDeciderTests: XCTestCase {
...
-                    status: .APPROVED,
-                    simProfiles: nil,
-                    kycStatusMap: KYCStatusMap(jumio: .PENDING, myInfo: .APPROVED, nricFin: .PENDING, addressPhone: .PENDING)
-                )
-            ]
+        let region = RegionResponse(
+            region: Region(id: "sg", name: "Singapore"),
+            status: .APPROVED,
+            simProfiles: nil,
+            kycStatusMap: KYCStatusMap(jumio: .PENDING, myInfo: .APPROVED, nricFin: .PENDING, addressPhone: .PENDING)
```
la3lma: This snippet of code (from liner 127 to 131) is repeated multiple times.  Is there a reason why it is not delegated to a method?

ostelco-core/Models/StageDecider.swift
```diff
@@ -40,12 +38,10 @@ public class LocalContext {
...
-        self.selectedRegion = selectedRegion
+    public init(hasFirebaseToken: Bool = false, hasAgreedToTerms: Bool = false, hasSeenNotificationPermissions: Bool = false, hasSeenVerifyIdentifyOnboarding: Bool = false, selectedVerificationOption: IdentityVerificationOption? = nil, myInfoCode: String? = nil, hasSeenESimOnboarding: Bool = false, hasSeenESIMInstructions: Bool = false, hasSeenAwesome: Bool = false, hasCompletedJumio: Bool = false, hasCompletedAddress: Bool = false, serverIsUnreachable: Bool = false, locationProblem: LocationProblem? = nil, hasSeenRegionOnboarding: Bool = false, hasSeenLocationPermissions: Bool = false, hasCameraProblem: Bool = false) {
```
la3lma: Stylistic question (I'm not an iOS developer): Is it common to have blocks like this without linebreaks and indentation in Swift?   At least in GitHub the code is a bit difficult to read quickly.

sgoodwin: Yes. That specific method would be a pain to read any way we did it because it takes so many parameters. Fortunately it's about to be greatly simplified.

ostelco-core/Models/StageDecider.swift
```diff
@@ -131,50 +111,60 @@ public struct StageDecider {
...
+        if localContext.hasFirebaseToken {
+            remove(.loginCarousel)
+        }
+        if localContext.hasAgreedToTerms {
+            remove(.legalStuff)
+        }
+        return stages[0]
+    }
+    
+    // swiftlint:disable:next cyclomatic_complexity
```
la3lma: What causes high cyclomatic complexity here?  There should be a comment explaining why the warning
is disabled.

dev-ostelco-ios-clientTests/CoordinatorTests/SingaporeUserHappyFlowWithSingPassStageDeciderTests.swift
sgoodwin: It's not always the same value for each, and embedding it into some shared value impedes the speed of working with tests. This way you can clearly see what values the test is considering and change them without chasing changes in other tests.

ostelco-core/Models/StageDecider.swift
sgoodwin: The method just below it has quite a few branching statements, that's what cyclomatic complexity is. The warning is disabled because this is the one place where we're allowing it since we know why.

dev-ostelco-ios-clientTests/CoordinatorTests/NorwayUserHappyFlowWithJumioStageDeciderTests.swift
```diff
@@ -12,66 +12,40 @@ import XCTest
...
-        XCTAssertEqual(decider.compute(context: context, localContext: localContext), .verifyIdentityOnboarding)
-    }
-    
-    func testUserHasSeenVerifyIdentifyOnboarding() {
-        let decider = StageDecider()
-        let localContext = LocalContext(selectedRegion: Region(id: "no", name: "NO"), hasSeenNotificationPermissions: true, regionVerified: true, hasSeenVerifyIdentifyOnboarding: true, hasSeenLocationPermissions: true)
-        let regions: [PrimeGQL.RegionDetailsFragment] = []
-        let context = Context(customer: CustomerModel(id: "xxx", name: "xxx", email: "xxxx@gmail.com", analyticsId: "xxxx", referralId: "xxxx"), regions: regions)
+        let localContext = LocalContext(hasSeenNotificationPermissions: true, hasSeenLocationPermissions: true)
+        let region = RegionResponse(
```
la3lma: This snippet of code (from line 16to 21) is repeated multiple times with small variations (.sg v.s. .no).  Is there a reason why it is not delegated to a method?

sgoodwin: Same reason as before.

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -49,23 +49,20 @@ class OnboardingCoordinator {
...
                 }
-            }.recover { error in
-                self.localContext.serverIsUnreachable = (error as NSError).code == -1004
-                let context: Context? = nil
-                let stage = self.stageDecider.compute(context: context, localContext: self.localContext)
-                self.afterDismissing {
-                    self.navigateTo(stage)
-                }
+        }.recover { error in
+            self.localContext.serverIsUnreachable = (error as NSError).code == -1004
```
la3lma: -1004 is a magic number.  What does it mean? Either add a comment, or declare a constant with an explanatory name.

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -250,26 +126,26 @@ extension OnboardingCoordinator: LoginDelegate {
     
     func signedIn(controller: UIViewController, authCode: String, contactEmail: String?) {
         let spinnerView = controller.showSpinner()
-
+
```
la3lma: Are these tabs?  Is there a policy on tabs v.s. spaces for swift code?

sgoodwin: We use what the editor does by default.

dev-ostelco-ios-clientTests/CoordinatorTests/EdgeCasesInStageDeciderTests.swift
```diff
@@ -17,22 +17,16 @@ class EdgeCasesInStageDeciderTests: XCTestCase {
...
-                        SimProfile(eSimActivationCode: "xxx", alias: "xxx", iccId: "xxx", status: .INSTALLED)
-                    ],
-                    kycStatusMap: KYCStatusMap(jumio: .PENDING, myInfo: .APPROVED, nricFin: .PENDING, addressPhone: .PENDING)
-                )
-            ]
+        let region = RegionResponse(
+            region: Region(id: "sg", name: "Singapore"),
+            status: .APPROVED,
+            simProfiles: [
+                SimProfile(eSimActivationCode: "xxx", alias: "xxx", iccId: "xxx", status: .INSTALLED)
```
la3lma: What does the "xxx" mean? Please add explanatory comment.

sgoodwin: It means the value doesn't matter. This is used all over the tests in places where the exact value doesn't matter. Not going to add a comment for all such values.

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
sgoodwin: -1004 means the server is unreachable. That's why we change `serverIsUnreachable` to true when the code equals that number.

PR #358, opened 2019-10-03
Phase 2!

sgoodwin:
This sets up the new tab order where the middle tab is now Coverage.
Havard is working on the much nicer looking UI that will be used here,
but this allows you to see the general idea. You can now go through
onboarding for singapore and norway separately. This selection of
countries is fairly arbitrary to show that onboarding can be summoned
for any region we expect the server to support.


dev-ostelco-ios-clientTests/CoordinatorTests/NorwayUserHappyFlowWithJumioStageDeciderTests.swift
```diff
@@ -25,7 +25,7 @@ class NorwayUserHappyFlowWithJumioStageDeciderTests: XCTestCase {
     
     func testUserHasCompletedJumio() {
         let decider = StageDecider()
-        let localContext = LocalContext(hasSeenNotificationPermissions: true, hasSeenVerifyIdentifyOnboarding: true, hasCompletedJumio: true, hasSeenLocationPermissions: true)
+        let localContext = LocalContext(hasSeenNotificationPermissions: true, hasCompletedJumio: true, hasSeenLocationPermissions: true)
```
la3lma: Much repeated code her.  Is it possible to delegate this either to a method or to the fixture setup for the test suite?

ostelco-core/Models/StageDecider.swift
```diff
@@ -38,11 +37,10 @@ public class LocalContext {
...
+    public init(hasFirebaseToken: Bool = false, hasAgreedToTerms: Bool = false, hasSeenNotificationPermissions: Bool = false, selectedVerificationOption: IdentityVerificationOption? = nil, myInfoCode: String? = nil, hasSeenESimOnboarding: Bool = false, hasSeenESIMInstructions: Bool = false, hasSeenAwesome: Bool = false, hasCompletedJumio: Bool = false, hasCompletedAddress: Bool = false, serverIsUnreachable: Bool = false, locationProblem: LocationProblem? = nil, hasSeenRegionOnboarding: Bool = false, hasSeenLocationPermissions: Bool = false, hasCameraProblem: Bool = false) {
```
la3lma: Very long line. Hard to read.  I prefer to have each parameter on a line, and any parameters (if any) that are worthy of special attention pointed out in comments.

ostelco-ios-client/ViewControllers/Country/CoverageViewController.swift
```diff
@@ -0,0 +1,72 @@
+//
+//  CoverageViewController.swift
+//  ostelco-ios-client
+//
+//  Created by Samuel Goodwin on 10/1/19.
+//  Copyright © 2019 mac. All rights reserved.
```
la3lma: This isn't strictly true :-). The file headers should reflect that this is red otter software.

havard024: our IDE (xcode) creates this, should be configurable though (I assume). I'm creating a task in notion to remember this.

dev-ostelco-ios-clientTests/CoordinatorTests/NorwayUserHappyFlowWithJumioStageDeciderTests.swift
havard024: Created a task in notion to address this

ostelco-core/Models/StageDecider.swift
sgoodwin: Yes we've discussed this already. This object is about to be simplified so it won't be quite so absurd.

dev-ostelco-ios-clientTests/CoordinatorTests/NorwayUserHappyFlowWithJumioStageDeciderTests.swift
sgoodwin: This isn't repeated code, it's just the data the test uses. There's no logic being duplicated here. Fixtures would only make this even more complicated.

ostelco-core/Models/StageDecider.swift
la3lma: (also I have very little object permanence when it comes to this codebase.  I mostly recognise only what I see in front of me at the moment, if it connects to or even if something I've seen a few days ago I will just not notice :-).

sgoodwin: Ok, but if you keep making the same comments and I keep having to respond to explain why it's not an issue, that's a bit frustrating for me. So maybe you could stop?

PR #361, opened 2019-10-08
First iteration fo multi region design

havard024:
Lists all countries in the design under coverage tab.

Simulator Screen Shot - iPhone Xʀ - 2019-10-08 at 12 07 21

Missing 
- Design for regions
- Logic to handle which countries / regions to show
- Logic to handle context button (GET, UPDATE, CURRENTLY ACTIVE)
- Clean up / restructure / refactor


sgoodwin:
Looks good! I guess next step would be hooking this up to that api change?

havard024:
Yeah, all that logic needs to be added

PR #366, opened 2019-10-09
Feature/add/region design to multi region feature

havard024:
see #364 for details


OstelcoStyles/OstelcoColor.swift
```diff
@@ -11,13 +11,15 @@ import SwiftUI
 
 /// Colors from https://app.zeplin.io/project/5c8b989f46989524fb0258ac/styleguide
 public enum OstelcoColor: String, CaseIterable {
+    case azul
```
sgoodwin, line 14: Are these the color names florian wanted? I thought we were using use-based names rather than color names?

havard024: There was no name for this color in zeplin, so I asked zeplin to name it for me, can update the name when it comes from design into zeplin

bloomingbridges: Hi guys, 
I'm still in the process of updating the Prod project. So far the only project that uses the new Component Library (and by extension has the most up-to-date styleguide with the right colours) is the OYA MultiSIM UI project.

sgoodwin (approved):

havard024:
RegionOnboardingCoordinator logic is not working as expected, thus preventing users from completing adding an ESim.

PR #367, opened 2019-10-09
Ask for permissions Before Home.

sgoodwin:
Because users can now choose to only grant permissions one time, we'll
need to be able to ask even after onboarding is complete.

### DIFF
--- a/ostelco-ios-client/ViewControllers/AuthParentViewController.swift
+++ b/ostelco-ios-client/ViewControllers/AuthParentViewController.swift
@@ -40,3 +40,4 @@ class AuthParentViewController: UIViewController, OnboardingCoordinatorDelegate 
         self.onboarding = onboarding
     }
 }
+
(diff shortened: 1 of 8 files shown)


ostelco-ios-client/ViewControllers/AuthParentViewController.swift
```diff
@@ -29,6 +30,15 @@ class AuthParentViewController: UIViewController, OnboardingCoordinatorDelegate
         
         let tabs = UIStoryboard(name: "TabController", bundle: nil).instantiateInitialViewController()
         embedFullViewChild(tabs!)
+        
+        let controller = LocationController.shared
+        if controller.authorizationStatus == .notDetermined {
```
havard024: What happens if `authorizationStatus == .denied`?

sgoodwin: If they refuse, then we can't try to be helpful, right?

havard024: I was under the impression that we would block the app if permission is denied, so they won't see the home screen. Exactly the same functionality for location permissions we have in AllowLocationPermissionsVC that we used in the onboarding flow (after user has logged in), move that logic out of the onboarding, but make sure it's handled all over the app

sgoodwin: You're right, I confirmed. Fixed now.

havard024 (approved):
👍

PR #371, opened 2019-10-15
Feature/convert settings to swiftui

havard024:
- PurchaseHistoryView is not complete, waiting for final design (will be fixed in a later in a separate PR)
- PurchaseHistoryView is presented using a modal instead of navigation, when I tried to use NavigationView + NavigationLink the back button ended up top left of the screen and it was unable to click on it. Not sure if this is a bug or if this is related to us embedding SwiftUI inside a ViewController.
- Show unread freshchat messages count on SettingsView


ostelco-ios-client/Views/SettingsView.swift
```diff
@@ -0,0 +1,138 @@
...
+                    }
+                    Divider()
+                    Button(action: {
+                        FreshchatManager.shared.show(self.controller)
+                    }) {
+                        HStack {
+                            OstelcoText(label: "Chat to Support")
+                            Spacer()
+                            Image(systemName: "chevron.right")
+                        }
```
havard024: Forgot to add badge for unread freshchat messages, working on that now

prasanthu (approved):

ostelco-ios-client/Views/SettingsView.swift
havard024: Fixed

PR #386, opened 2019-10-22
Add select verification method option before jumio.

havard024:
Also made SelectIdentityVerificationMethodViewController radio buttons render based on the list of IdentityVerificationOption available.

Missing:
SelectIdentityVerificationMethodViewController has to be styled according to design.


sgoodwin:
I thought we weren't going to have them pick an option when there wasn't a choice?

PR #393, opened 2019-10-23
Fix the location check for Singapore

sgoodwin:
The app was allowing random people to try to sign up in Singapore. This
is no good!

### DIFF
--- a/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
+++ b/ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
@@ -118,15 +118,6 @@ class OnboardingCoordinator {
             navigationController.setViewControllers([awesome], animated: true)
         }
     }
-    
-    private func checkLocation(country: Country) -> LocationProblem? {
-        let controller = LocationController.shared
-        if controller.checkInCorrectCountry(country) {
-            return nil
-        } else {
-            return controller.locationProblem
-        }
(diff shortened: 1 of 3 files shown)


havard024:
We want to hold off merging this until after the weekend, that we don't check location for Singapore right now is actually a good thing, it enables our testers to try out myinfo and scanIC.

havard024 (approved):
👍

havard024:
There was a merge conflict related to the RegionOnboardingCoordinator, tried to resolve it so we can merge this PR.

ostelco-ios-client/Coordinators/OnboardingCoordinator.swift
```diff
@@ -512,22 +503,34 @@ class RegionOnboardingCoordinator {
...
-            .first(where: {
+            let region = context.regions.map {
+                $0.fragments.regionDetailsFragment
+            }.first(where: {
                 $0.region.id == self.region.region.id
             })!
             
-            let stage = self.stageDecider.stageForRegion(region: RegionResponse(gqlData: region), localContext: self.localContext)
+            let stage: StageDecider.RegionStage
+            if let problem = self.checkLocation(country: Country(region.region.id)) {
```
havard024, line 529: @sgoodwin regionID is not a country code, we can't use region id to create countries, this will fail in the future

PR #394, opened 2019-10-23
Fix issue with parsing incoming push notifications

sgoodwin:
I think because of an iOS change, push notification data can come in
slightly differently. This would result in the app crashing, so we need
to account for this. This will make the app check both and not get
confused as easily if the payload is formatted differently.


havard024 (approved):
👍

PR #395, opened 2019-10-23
Fix jittering issues when becoming active.

sgoodwin:
This prevents the tab bar jumping if the app is paused at all, this also
prevents the in-progress EKYC from being dismissed when the app is
paused as well. This doesn't completely eliminate the mild visual
flickering when the app is un-paused.


havard024 (approved):
👍

PR #398, opened 2019-10-24
Add the address step to all jumio flows.

sgoodwin:
The address step cannot be skipped for people doing Jumio outside of
Singapore. It is a requirement from the government that we still show
the address step for people in Malaysia for example.


vihangpatil (approved):
LGTM

PR #400, opened 2019-10-24
Fix Onboarding from Shortcuts

sgoodwin:
Since there's special case shortcuts to ekyc for singapore and malaysia
for everyone in Norway, it resulted in a case where they would try to do
ekyc and wait forever to be approved. In the system we could see they
*were* approved, but for Norway (even when that's not a region they
could select). This fixes RegionOnboardingCoordinator to only take in
region to avoid such confusion later.


ostelco-ios-client/Coordinators/EKYC/JumioCoordinator.swift
```diff
@@ -92,7 +90,7 @@ class JumioCoordinator: NSObject {
         config.delegate = self
         
         // This must be a 3-letter code following ISO-3166
-        config.preselectedCountry = country.threeLetterCountryCode
+        config.preselectedCountry = Country(regionID).threeLetterCountryCode
```
havard024, line 93: @sgoodwin this is not going to fly, regionID is not equal to country code, this has to be reverted else we can get issues in the future, when we get a region with ID that does not resemble a country code.

PR #402, opened 2019-10-25
Show welcome to country message if user changes country.

havard024:
- Create a GlobalStore with the purpose of sharing common store logic accross views
- Move current country from CoverageStore to GlobalStore since it's now needed for both BalanceView and CoverageView
- Add logic to cache current country so we can use it to compare if user changes country
- Create a component out of the message view so it can be reused for different types of balance screen messages

### DIFF
--- a/ostelco-ios-client/Features/App/GlobalStore.swift
+++ b/ostelco-ios-client/Features/App/GlobalStore.swift
@@ -61,4 +61,33 @@ final class GlobalStore: ObservableObject {
         }
         return nil
     }
+    
+    func allowedCountries() -> [String] {
+        let countryCodeToRegionCodeMap = RemoteConfigManager.shared.countryCodeAndRegionCodes.reduce(into: [:], { (result, value) in
+            result[value.countryCode] = value.regionCodes
+        })
+        
+        if let regionDetailslist = regions {
+            let regionCodes = Set(
+                regionDetailslist.map({ $0.region.id })
+            )
+            
+            let allowedCountryCodes = Array(
+                countryCodeToRegionCodeMap
+                    .filter({ (_, value) in Set(value).intersection(regionCodes).isNotEmpty })
+                    .keys
+                )
+            
+            return allowedCountryCodes
+        }
+        
+        return []
+    }
+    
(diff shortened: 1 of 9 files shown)


ostelco-ios-client/Features/App/GlobalStore.swift
```diff
@@ -0,0 +1,64 @@
...
+                .filter({ regionCodes.contains($0.region.id) })
+                .filter({ $0.simProfiles != nil })
+                .map({ $0.simProfiles!.map({ $0.fragments.simProfileFields }) })
+                .reduce([], +)
+        }
+        return []
+    }
+    
+    func showCountryChangedMessage() -> Country? {
+        if let previousCountry = previousCountry, let country = country {
```
sgoodwin, line 55: Shouldn't the thing already meant to monitor the current country be the one responsible for determining if it changed?

havard024: It is here:
NotificationCenter.default.addObserver(self, selector: #selector(countryChanged(_:)), name: CurrentCountryChanged, object: nil)
then the value is stored in the store as a Published variable which will trigger the SwiftUI view to update

sgoodwin: The store appears to be caching the value to decide if the country has changed.

havard024: Since the store is responsible for updating the UI the changes has to go through the store. That also means the store has to listen for the country changes itself.
That's at least my current understanding of SwiftUI and view updates, if you have a better idea on how to solve this, I'm all ears

sgoodwin: I’m saying the logic to decide when the country is different should not be here. This should be able to be notified of the change and be done. Now the job of keeping track of countries has been spread between two entities.

havard024: I'm not sure how you want this to be, it would probably be easier if you do it, to show me what you are thinking about

sgoodwin (approved):

PR #406, opened 2019-10-27
Revert "Handle being forced out better"

havard024:
Reverts ostelco/ostelco-ios-client#401

The PR prevents new users from using the app, when a new user logs in using sign in with apple, we don't have a user in our servers yet, thus the UserNotFound error is triggered and the app logs the user out.


havard024:
#407

PR #417, opened 2019-10-30
Add scripts to update remote config via REST APIs.

havard024:
An attempt to document the use of remote config as well as to automate the process of updating remote config without modifying the data in firebase console directly.


config/README.md
```diff
@@ -1,6 +1,67 @@
-We use [Firebase Remote Config](https://firebase.google.com/docs/remote-config) for dynamic configuration of our app.
+This project is based on [REST APIs using the node client](https://github.com/firebase/quickstart-nodejs/tree/master/config) to manipulate remote config in firebase.
+
+More information about what we are using remote config for can be found in this [private gdrive document](https://docs.google.com/document/d/1UNAAbFE_U8KVSd5SNg_Z0Gm9Kn7TY2UbQbbsDdz6ZGg)
+
+The project is written using our favourite command / programming languages `javascript` and `bash`, though fear not, the project can easily be converted to other languages, here are some other quickstart examples of firebase remote config:
```
la3lma, line 5: Iiik :-).  I'm currently trying to avoid using both, by using Golang instead :-).    That said.  The general idea of using a proper programming language (and I will grudgingly admit that .js is a proper programming language) instead of just a scripting language like Bash is a practice we should encourage and evolve.

config/README.md
```diff
@@ -1,6 +1,67 @@
...
+
+```
+node -v
+v10.15.3
+npm -v
+6.11.3
+```
+
+### Prerequisites
+1. Download the corresponding service account files and add them to .secrets
```
la3lma: .secrets where? Root directory of the project, your user, or what? Be specific.

config/index.js
```diff
@@ -0,0 +1,179 @@
...
+var ENVIRONMENT = process.env.ENVIRONMENT || 'dev'
+var PROJECT_ID = `pi-ostelco-${ENVIRONMENT}`;
+var HOST = 'https://firebaseremoteconfig.googleapis.com';
+var PATH = '/v1/projects/' + PROJECT_ID + '/remoteConfig';
+var SCOPES = ['https://www.googleapis.com/auth/firebase.remoteconfig'];
+
+/**
+ * Get a valid access token.
+ */
+// [START retrieve_access_token]
```
la3lma, line 14: I'm unfamiliar with .js coding conventions.   Does the "// [START ...]" pattern have some special significance ?

config/index.js
```diff
@@ -0,0 +1,179 @@
+var rp = require('request-promise');
+var fs = require('fs');
+var google = require('googleapis');
+
+var ENVIRONMENT = process.env.ENVIRONMENT || 'dev'
+var PROJECT_ID = `pi-ostelco-${ENVIRONMENT}`;
+var HOST = 'https://firebaseremoteconfig.googleapis.com';
+var PATH = '/v1/projects/' + PROJECT_ID + '/remoteConfig';
```
la3lma, line 8: Where is this path pointing? I'm unfamiliar with the /v1/projects .... , is that a coordinate inside firebase's URL space?

config/index.js
```diff
@@ -0,0 +1,179 @@
...
+  listVersions();
+} else if (action && action === 'rollback' && etagOrVersion) {
+  rollback(etagOrVersion);
+} else {
+  console.log('Invalid command. Please use one of the following:\n'
+      + 'node index.js get\n'
+      + 'node index.js publish <LATEST_ETAG>\n'
+      + 'node index.js versions\n'
+      + 'node index.js rollback <TEMPLATE_VERSION_NUMBER>');
+}
```
la3lma: Add a final newline (nitpicking).

config/index.js
```diff
@@ -0,0 +1,179 @@
...
+          console.error('Unable to publish template.');
+          console.error(err);
+        });
+  });
+}
+
+const action = process.argv[2];
+const etagOrVersion = process.argv[3];
+
+if (action && action === 'get') {
```
la3lma, line 165: Looks a tad messy.  Is it possible to use a switch instead (if the answer isn't obvious in ten seconds to an experienced .js hand as you it probably isn't easy and the existing solution should be used instead)

config/update.sh
```diff
@@ -0,0 +1,4 @@
+# Get previous ETAG
```
la3lma: Don't know how this is intended to be used for, but I suggest the following improvements
1) Add a shebang, so that we know which shell this is intended to be parsed by (even if it si to be sourced, that is helpful)

2) Add a brief (one or two lines) of comments at the top stating the usage of the script.  I see that $1 is being used further down, but it's hard to determine what that's ment to be from just staring at the code.   Bash is concise, so some level of documentation is necessary

3) If. this is bash, use $(..) instead of `...` to get results from sub shells

4) The ETAG=... line is fair complex, so an one (or up to three)-line comment above it would help
     comprehension a lot.

5) Not necessary, but a practice I generally try to follow, so I'll mention it for you:  You use several subcommands (node, grep, cut, sed) that are usually, but not always installed.   I find it to be a good practice to have list of necessary subcommands, and then check if they can be found, and fail early if not.   In this particular script I actually don't know what the result would be, but in general the benefit is that the script fails early, rather than chugging through some process, and perhaps leaving something in an intermediate state, and then failing because a dependency can't be found.  It is much better to fail early with a clear statement about what is missing, and no intermediate state to clean up.

config/remote-config-prod.json
```diff
@@ -1,5 +1,28 @@
...
+      "defaultValue": {
+        "value": "[{\"countryCode\":\"AU\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"CN\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"HK\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"IN\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"ID\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"JP\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"PH\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"KR\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"TH\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"VN\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"SG\",\"regionCodes\":[\"sg\"]},{\"countryCode\":\"US\",\"regionCodes\":[\"us\"]},{\"countryCode\":\"MY\",\"regionCodes\":[\"my\"]},{\"countryCode\":\"NO\",\"regionCodes\":[\"no\"]}]"
+      }
+    }
+  },
+  "version": {
+    "versionNumber": "3",
+    "updateTime": "2019-10-29T13:07:48.179Z",
+    "updateUser": {
+      "email": "havard@redotter.sg"
```
la3lma: Could this somehow be substituted in from somewhere more prominent?  Adding personal email addresses is in general a discouraged practice, but when it has to be done, it should be done in a way that is prominent, and easy to modify when the opportunity arises.

config/remote-config-dev.json
```diff
@@ -1,5 +1,28 @@
...
+      "defaultValue": {
+        "value": "{\"enableStripeInsteadOfApplePay\":false}"
+      }
+    }
+  },
+  "version": {
+    "versionNumber": "37",
+    "updateTime": "2019-10-30T12:40:37.554Z",
+    "updateUser": {
+      "email": "firebase-adminsdk-robwm@pi-ostelco-dev.iam.gserviceaccount.com"
```
la3lma, line 23: Who is this user?   Some firebase admin user, or one of ours?

la3lma (approved):

config/index.js
havard024: PATH is used when calling the APIs, HOST + PATH = API and in some cases HOST + PATH + ':listVersions?pageSize=5'
There's nothing special and could've been done in many different ways.

config/index.js
havard024: I'm also unfamiliar with this type of syntax, it was part of the template and I didn't want to heavily modify the quickstart template provided by firebase.

config/index.js
havard024: Switch is absolutely possible, but ref earlier comment, this is based on a [template from firebase](https://github.com/firebase/quickstart-nodejs/tree/master/config), so I didn't want to modify it too much.

config/remote-config-dev.json
havard024: My best guess would be that this is the author of the update, and the author is based on the service account from firebase.

havard024: This exists in the service-account.json under `client_email`

config/README.md
havard024: I've added a step describing that you need to create a folder .secrets in the current folder.

config/remote-config-prod.json
havard024: I agree, this email most likely came from when I changed the remote config manually through firebase console. If we change it using this script, the correct email should be the `client_email` inside service-account.json for production.  I've updated the email accordingly.

config/README.md
havard024: I agree, there is no need to use both, it was more of an convenience, luckily there are other language options out there, in case someone's fingers are itching to convert :D

config/update.sh
havard024: 1. Thanks, forgot 👍 
2. Added
3. Wrapped the $1 in "$1" though the values are only supposed to be prod and dev and the script is supposed to be run through npm using `npm run deploy:dev` or `npm run deploy:prod` as described in the `README`. I've made a comment about it in the `update.sh` to be more clear about it. Also, ideally, you would also validate the values of $1 so you can't use anything else.
4. I've added a comment describing what it does.
5. I agree, node and npm are already listed as dependencies in the `README`, could probably be clearer, but they are listed under `Environment`. Also agree that it would be great to check what tools are available etc, but it felt like too much work for a not so big task, also each step can be run manually and the script will probably be rewritten at some point. Thus as long as there are enough comments to give comprehensions of what's going it.

PR #432, opened 2019-11-18
Fix the issue with showing welcome message late.

sgoodwin:
Since we don't know on the server end that a sim is installed right away
(which backend is working on), it makes more sense to stop showing the
welcome message once the user has finished eKYC instead. By that time,
they likely know what's going on and can re-download the sim if they
need to.


la3lma (approved):
lgtm

PR #438, opened 2019-11-25
Appearance Cleanup

sgoodwin:
Attempting to correct some weird display issues by modernizing the way we adjust overall appearance.


ostelco-ios-client/SupportingFiles/Production.plist
```diff
@@ -28,7 +28,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>16</string>
+	<string>21</string>
 	<key>FacebookAppID</key>
 	<string>2328394300810303</string>
```
la3lma, line 33: Is this number in any way something we should keep secret or is this public information?

sgoodwin: It's publicly viewable.

PR #439, opened 2019-11-26
Report successfully installing eSims

sgoodwin:
The backend is not always informed in a timely fashion, so this allows
the app to report that the eSim was successfully installed instead. The
app will look for either this report or the overal status from the
backend to decide if a sim is setup.


vihangpatil (approved):
Just glanced through the change.

PR #440, opened 2019-11-28
Add a caution for users.

sgoodwin:
If you are signing up in a country that is not your current country,
we'll show a warning to explain what's going on. Hopefully this will
help to avoid confusion from users.


sgoodwin:
Simulator Screen Shot - iPhone 11 Pro Max - 2019-11-28 at 13 48 25

